PR #1, opened 2021-09-03
Adding initial implementation of Sentry integration for edX

blarghmatey:
- Adds an app entrypoint for lms and cms
- Adds a settings module that sets up the Sentry integration with sane defaults

### DIFF
--- a/src/ol_openedx_sentry/BUILD
+++ b/src/ol_openedx_sentry/BUILD
@@ -1,0 +1,19 @@
+python_library(
+    name="sentry_app",
+    dependencies=["src/ol_openedx_sentry/settings:sentry_settings"]
+)
+
+python_distribution(
+    name="sentry_package",
+    dependencies=[":sentry_app", "src/ol_openedx_sentry/settings:sentry_settings"],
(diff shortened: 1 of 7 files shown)


rhysyngsun (approved):
generally looks good to me

src/ol_openedx_sentry/BUILD
```diff
@@ -0,0 +1,13 @@
+python_distribution(
+    name="sentry",
```
shaidar, line 2: Is this right?

src/ol_openedx_sentry/app.py
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+
+class EdxSentryCMS(AppConfig):
+    name = "ol_openedx_sentry"
+    verbose_name = "Sentry integration for Open edX"
+    plugin_app = {
+        "settings_config": {
+            "cms.djangoapp": sentry_settings_config,
+        }
```
shaidar, line 29: Why the difference in the plugin_app between LMS and CMS?

shaidar:
👍

PR #2, opened 2021-09-08
Adding initial pass of extending logging configuration for edx-platform

blarghmatey:
- Adds a json formatter for emitting structured log messages
- Adds a rotating file handler for writing JSON structured log messages to disk

### DIFF
--- a/src/ol_openedx_logging/settings/production.py
+++ b/src/ol_openedx_logging/settings/production.py
@@ -1,0 +1,46 @@
+from typing import Any, Dict
+
+MEGABYTE = 1024 * 1024
+
+
+def _load_env_tokens(edx_settings, default_settings: Dict[str, Any]) -> Dict[str, Any]:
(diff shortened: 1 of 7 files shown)


src/ol_openedx_logging/settings/production.py
```diff
@@ -0,0 +1,46 @@
+from typing import Any, Dict
+
+MEGABYTE = 1024 * 1024
+
+
+def _load_env_tokens(edx_settings, default_settings: Dict[str, Any]) -> Dict[str, Any]:
```
shaidar, line 6: What are `edx_settings` and where are those defined?

blarghmatey: That's a parameter that gets passed into the `plugin_settings` function which then propagates to the `_load_env_tokens` function to populate the parameters that we care about in our plugin. In practice it's the Django settings module object after it's modified by e.g. `production.py`

shaidar:
👍

Issue #4, opened 2021-10-29
Port canvas_integration django app from mitodl/edx-platform


arslanashraf7:
Subtask of #3 

Depends on #7, #6 

**What needs to be done?**
This task will include the effort to port all the Canvas Integration-related code that should/can be moved into a plugin.

The plugin should be created in a way to be supported by edX platform itself. For creating edX extension plugins a good read could be [this guide](https://github.com/edx/edx-platform/blob/master/docs/guides/extension_points.rst) and this [utility repo](https://github.com/edx/edx-django-utils). Following the edX standards, we can make sure that our plugins work with edX without any issues.

### Acceptance Criteria:

- [x] Migrate all the canvas_integration Django app code into the `open-edx-plugins` inside `/src` directory (Parallelly this part will be removed from `edx-platform` as part of #6)
- [x] Setup all the configurations in the app to be pip installable within edX-platform
- [x] Setup the BUILD files to generate proper app releases for migrated `canvas_integration` Django app

arslanashraf7:
Closed via #8

PR #8, opened 2021-11-04
feat: add canvas integration plugin

arslanashraf7:
### Related Tickets:
https://github.com/mitodl/open-edx-plugins/issues/4

### What's this PR does?
- Migrates the `canvas_integration` app from `edx-platform` into a separate plugin

### How to test?
To test this PR you will need a counterpart in the `edx-platform` and that part is specifically done in https://github.com/mitodl/edx-platform/pull/274

- Setup your edx-platform on the associated branch https://github.com/mitodl/edx-platform/pull/274 from here
- Install this plugin (e.g. generate a build for this plugin with `./pants package ::` and then install that into the edx-platform)
- Restart the lms (`make lms-restart`)
- Make sure to set the values for CANVAS_ACCESS_TOKEN=?, CANVAS_BASE_URL=https://mit.test.instructure.com` (You can ask me for the Token value)
(More instruction to be added once the PR is finalized)
- Now open a course as admin and click `Instructor` tab and make sure you see a `Canvas` tab under it
- Click the canvas tab and test all the functionality of canvas integration (e.g listing enrollments, syncing enrollments, pushing grades)

### Background context:
We had made several changes inside our for of the edx-platform where we added a Django application to support canvas integration. In the process, we made many changes in open edX's `Instructor` django application since we wanted to run `canvas_integration` inside the `Instrutor` tab. (More details to come)

### Screenshots:
**Landing page and tasks details**
screencapture-localhost-18000-courses-course-v1-edX-DemoX-Demo-Course-instructor-2021-11-12-15_13_09

**List Canvas Enrollments**
Screenshot 2021-11-12 at 3 14 15 PM

**List Canvas Assignments**
Screenshot 2021-11-12 at 3 14 33 PM


src/ol_openedx_canvas_integration/api.py
```diff
@@ -0,0 +1,220 @@
...
+                result = 'Enrolled user in the course'
+            except Exception as ex:  # pylint: disable=broad-except
+                result = 'Failed to enroll - {}'.format(ex)
+        else:
+            result = 'User already enrolled'
+        results[email] = result
+    return results
+
+
+def get_subsection_user_grades(course):
```
ziafazal, line 70: @arslanashraf7 I think we can remove a lot of code around getting user's grate by assignment and use builtin- [method](https://github.com/edx/edx-platform/blob/3152c14192dcccac79b7b7a76eba91b380eba78a/lms/djangoapps/grades/models_api.py#L78) of PersistentGrades

src/ol_openedx_canvas_integration/client.py
```diff
@@ -0,0 +1,204 @@
...
+        self.canvas_course_id = canvas_course_id
+
+    @staticmethod
+    def get_canvas_session():
+        """
+        Create a request session with the access token
+        """
+        session = requests.Session()
+        session.headers.update({
+            "Authorization": "Bearer {token}".format(token=settings.CANVAS_ACCESS_TOKEN)
```
ziafazal, line 26: I think we can persist canvas credentials in encrypted field of django model instead of using them from settings

src/ol_openedx_canvas_integration/client.py
```diff
@@ -0,0 +1,204 @@
...
+    Returns:
+        dict:
+            Assignment payload to be sent to Canvas to create or update the assignment
+    """
+    return {
+        "assignment": {
+            "name": subsection_block.display_name,
+            "integration_id": str(subsection_block.location),
+            "grading_type": "percent",
+            "points_possible": DEFAULT_ASSIGNMENT_POINTS,
```
ziafazal, line 176: possible grade of actual subsection instead of default value.

src/ol_openedx_canvas_integration/settings/common.py
```diff
@@ -0,0 +1,8 @@
+"""Common settings unique to the canvas integration plugin."""
+
+
+def plugin_settings(settings):
+    """Settings for the canvas integration plugin."""
+    settings.CANVAS_ACCESS_TOKEN = None
+    settings.CANVAS_BASE_URL = None
+    settings.FEATURES['ENABLE_CANVAS_INTEGRATION'] = False
```
ziafazal, line 8: By default this feature should be enabled.

ziafazal (commented):
@arslanashraf7 few improvements I suggested which are not necessarily needed to be done right now but would be useful in future.

src/ol_openedx_canvas_integration/client.py
pdpinch: @arslanashraf7 can you check with @mitodl/devops for their preference for managing the secrets?

src/ol_openedx_canvas_integration/settings/common.py
pdpinch: I agree with the default, but do we even need a feature flag? If the plugin is installed, shouldn't that be a clear signal that it should be turned on?

src/ol_openedx_canvas_integration/client.py
arslanashraf7: Sure, Currently we are injecting them from the vault https://github.com/search?q=org%3Amitodl+CANVAS_ACCESS_TOKEN&type=code. I'll talk to the ops guys if they want to go with this approach or the existing vault approach is better.

src/ol_openedx_canvas_integration/settings/common.py
arslanashraf7: We've been running the canvas behind a feature flag from when we first implemented that & I'm not sure what was the reason to keep it behind a flag but we can remove it and assume if the plugin is installed, the canvas is enabled.
The flag is used only once in the [edx counterpart](https://github.com/mitodl/edx-platform/pull/274/files#diff-0d1ee66fb84bee6c3517980a7cefd74a050ccb60da9c0c3b662ab0a2d80d17c0R151) and we can make that check `settings.INSTALLED_APPS` instead of a flag. I'll remove the feature flag now unless you say otherwise.

pdpinch: I think it would be better to test on INSTALLED_APPS, but if you want to leave that for another issue & PR, I would be OK with that.

src/ol_openedx_canvas_integration/client.py
blarghmatey: It's fine to keep it as a settings attribute because we can take the same approach of managing the value through the configuration template.

src/ol_openedx_canvas_integration/api.py
arslanashraf7: As discussed keeping the scope of this in mind, I've created a ticket to keep track of this optimization https://github.com/mitodl/open-edx-plugins/issues/10.

src/ol_openedx_canvas_integration/client.py
arslanashraf7: This would be done as part of https://github.com/mitodl/open-edx-plugins/issues/10 since it will be easier to directly get the max weightage/score for a subsection.

src/ol_openedx_canvas_integration/settings/common.py
arslanashraf7: Updated.

arslanashraf7:
@ziafazal I've addressed feedback and responded to some comments. This is ready for review again, Could you please take a look.

ziafazal (approved):
👍

gsidebo (approved):
👍 

This PR has been sitting for a while since it has taken so long for me to address other issues, get `pants` to actually build the plugin, and get the plugin itself working in devstack. It seems that my Canvas token is outdated, so I didn't directly test the functionality, but I finally managed to get the Canvas tab to show up and work correctly. Given the timing I think it's best to merge this as it is, and if there are any lingering bugs, those can be addressed in future PRs. I have a couple concerns though:

1. The instructions on this PR are not complete. I had to dig around past issues/PRs to figure out that I needed to add a `canvas_course_id` setting to my course, which I ended up doing in Studio. That leads me to the next concern...
2. Why is the documentation not included in this PR? Why would we add it in a future PR? This is the PR that adds the feature, so why not merge it with instructions to use the feature?

Anyway, feel free to merge this as it is and please keep those notes in mind for future PRs

arslanashraf7:
@gsidebo I'm sorry I somehow missed some details in the PR description but thanks for reviewing this PR anyway. I agree with your idea of adding the documentation along with the feature which should be followed in the oncoming PRs. 
For the documentation of this plugin, I'll create a follow-up PR.

Issue #15, opened 2021-11-15
ol_openedx_sentry :update for Maple release


pdpinch:
Test the ol_openedx_sentry plugin with the maple release of openedX 

In particular, it may need updates for the django 3.2 upgrade.

blarghmatey:
We are already using this plugin on our Maple builds.

arslanashraf7:
Closed via #20

Issue #14, opened 2021-11-15
ol_openedx_logging: update for Maple release


pdpinch:
Test the ol_openedx_logging plugin with the maple release of openedX 

In particular, it may need updates for the django 3.2 upgrade.

blarghmatey:
We are already using this plugin on our Maple builds.

arslanashraf7:
Closed via #20. Can be reopened if we see any issues with the plugin.

PR #20, opened 2021-11-26
build: update the Django versions support(3.2)

arslanashraf7:
## Related Ticket:
#14 , #15 

## What this PR does:
- Tightens the boundary for the Django version to be used (`above 2.0 & below 4.0`) 
- Updates the poetry config

## How to test
- Generate new plugin builds for `ol-openedx-logging` & `ol-openedx-sentry` and install them into the Maple release of edX and make sure that things work fine.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -41,7 +41,7 @@ tests_no_zope = ["coverage[toml] (>=5.0.2)", "hypothesis", "pympler", "pytest (>
 
 [[package]]
 name = "backports.entry-points-selectable"
-version = "1.1.0"
+version = "1.1.1"
 description = "Compatibility shim providing selectable entry points for older implementations"
 category = "dev"
 optional = false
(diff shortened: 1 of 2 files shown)


arslanashraf7:
@blarghmatey When possible can you add your comment on below:

1- I wasn't able to install [systemd-python](https://github.com/mitodl/open-edx-plugins/pull/20/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R14) on the MACOS. So, in order to update the poetry, I removed `systemd-python` package, updated the poetry with `Django`, and then manually added ``systemd-python` configs back. (is it fine with you?)

2- To test the `ol-openedx-loggin` plugin I was able to see the JSON formatted logs details in `/var/log/edxapp/app.log` (Is there any more to testing this?)

3- Is it possible to test `ol-openedx-sentry` (What could be the way to test that? Probably just events filtering?)

blarghmatey:
So, the systemd-python dependency can be removed entirely as it is not being actively used anywhere due to complications around ensuring C library dependencies. As to testing the logging and Sentry packages, we are already using them in Maple and on MITx Online so there's not any further testing required. It's also worth noting that because we are using Pants for building the packages, pinning the Django requirement in the pyproject.toml file it won't have any real impact on the generated setup.py for those packages.

arslanashraf7:
@blarghmatey Thanks for adding the details. Although I tried to install the `systemd-python` within the edX container and i was able to do so with other dev dependency installations.

>So, the systemd-python dependency can be removed entirely as it is not being actively used anywhere 

But as you mentioned in your comment I'll remove this package.

> It's also worth noting that because we are using Pants for building the packages, pinning the Django requirement in the pyproject.toml file it won't have any real impact on the generated setup.py for those packages.

I tried with `./pants package ::` and checking the setup.py it turns out that the version requirements for Django do come from `pyproject.toml` but since we are already accepting `version >2.0` we shouldn't have any specific issues nor did I see any issues while testing that might need any updates regarding `Django3.2`. So, I kept it to `>2.0`

@blarghmatey can you can take a look at the PR now & let me know if it needs any changes.

blarghmatey:
Looks good :+1:

PR #24, opened 2021-12-03
fix: commit urls associated with platform

arslanashraf7:
## Related Tickets:
None

## What this PR does?
- Fixes the references in the READMEs.

## Why?
Our plugin PRs were merged first and then the platform counterparts which were squashed and the commit hashes changed with new commit. So, This PR just fixes that.

## How to test?
Just verify the links point to the correct URLs.


gsidebo (approved):
👍 👍

Issue #27, opened 2022-02-03
Create a plugin to provide a course export API


blarghmatey:
As part of our data pipelines we need to be able to export the course contents as OLX. In order to make this more manageable in a proper orchestration approach we want to create an API endpoint that takes a course ID as an input and exports the course to an S3 bucket. This API should respond with the location that the file will be exported to. As an optional enhancement, it would be useful to allow passing in a list of course IDs and exporting all of them to the same location.

There is already a majority of the code needed to support this use case.

Code to export a course tarball: https://github.com/edx/edx-platform/blob/open-release%2Fmaple.master/cms/djangoapps/contentstore/tasks.py#L332
An API view for course import: https://github.com/edx/edx-platform/blob/open-release%2Fmaple.master/cms/djangoapps/contentstore/api/views/course_import.py

blarghmatey:
The immediate use case is for being able to install this in our existing edX deployments to be consumed by our own internal service. As such we can just build the functionality, deploy and validate it, and add tests as a second step. Eventually we might want to contribute this functionality upstream, but that is not a requirement for the initial work.

arslanashraf7:
So the synchronous course upload was intended to see this being run inside a pipeline and look for the s3 bucket URLs as a response upon the API call. Mentioned also in the [Reviewer Note](https://github.com/mitodl/open-edx-plugins/pull/29#issue-1132118020).

**If we are looking to change the behavior:**
>Is there an available interface for defining a task inside a plugin?

Looking at a few references e.g. [Django App Plugins](https://github.com/openedx/edx-django-utils/tree/master/edx_django_utils/plugins), [Plugin Contexts](https://github.com/openedx/edx-django-utils/blob/master/edx_django_utils/plugins/docs/decisions/0001-plugin-contexts.rst) I don't think there is any interface right away as part of the plugin configuration to define a Task.

>Alternatively, how difficult will it be to defer the heavy lifting to Celery?

Looking at how [course import](https://github.com/openedx/edx-platform/blob/open-release%2Fmaple.master/cms/djangoapps/contentstore/api/views/course_import.py#L150-L154) is working, It is also starting celery task to perform the import operation for a single course only. And I think we might do something similar too.


>This API should respond with the location that the file will be exported to.

_**But,**_ As mentioned in the ticket description, This will change the behavior of the API itself, The course import view gives 2 APIs 
1) POST - for starting a celery task for a course ID and then returning the task id for future progress updates 
[...]

PR #29, opened 2022-02-11
feat: Add course export s3 plugin

arslanashraf7:
## Related Ticket:
#27 

## What this PR does?
- Add basic architecture of a django plugin for edX
- Adds a new API to generate OLX course exports and upload to s3

## How should this be tested?
-  Generate a build for the package. We are using 'Pants' for that. Run `./pants package ::` in the main directory in the terminal which will generate many `.tar.gz & .whl` packages for you.
- Run `make studio-shell` and install `ol-openedx-coruse-course.*` (.gz or .whl)
-  You need to set the configuration values for the S3/AWS. e.g. set proper values for `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY` and `COURSE_IMPORT_EXPORT_BUCKET`. (All of these can be set in `private.py`)
- You will need a django admin user to test this API otherwise you would not be authorized
- Check access by calling the API without `Authorization` header and notice `400 with authorization error message`
- Check access with `Authorization` header with a token of some non admin User and `notice 400 with the error message`
- Call the API without setting AWS settings or bucket name, You should see a `400 with the proper error message`
- Now call the API (`<STUDIO_BASE>/api/courses/v0/export/`) and POST course IDs like:
```
{
   "courses": ["course-v1:edX+DemoX+Demo_Course", "course-v1:edX+Test+Test_Course"]
}
```
- If everything is set properly you should be able to notice a `200` status code with response something like:
```
{
    "successful_uploads": {
        "course-v1:edX+DemoX+Demo_Course": "https://bucket_name.s3.amazonaws.com/course-v1:edX+DemoX+Demo_Course.tar.gz",
        "course-v1:edX+Test+Test_Course": "https://bucket_name.s3.amazonaws.com/course-v1:edX+Test+Test_Course.tar.gz"
    },
    "failed_uploads": {}
} 
```

- If `COURSE_IMPORT_EXPORT_BUCKET` is set to `courseexports`, That means your courses will be uploaded/exported to a bucket named `courseexports`.
- The API is authenticated and only permitted to admin users so that Access Token you use should belong to an admin user. (be mindful of this while hitting the API).

## Reviewer Note:
- The API uploads the courses synchronously, I decided to make it synchronous since it is going to be a part of a pipeline and we might look on waiting for the API to properly respond because we are looking for the uploaded location URLs for the posted cours IDs.

### DIFF
--- a/src/ol_openedx_course_export/LICENSE.txt
+++ b/src/ol_openedx_course_export/LICENSE.txt
@@ -1,0 +1,28 @@
+Copyright (C) 2022  MIT Open Learning
+
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright notice, this
+  list of conditions and the following disclaimer.
+
+* Redistributions in binary form must reproduce the above copyright notice,
+  this list of conditions and the following disclaimer in the documentation
+  and/or other materials provided with the distribution.
+
+* Neither the name of the copyright holder nor the names of its
+  contributors may be used to endorse or promote products derived from
+  this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
--- a/src/ol_openedx_course_export/utils.py
+++ b/src/ol_openedx_course_export/utils.py
@@ -1,0 +1,36 @@
+from django.conf import settings
+
+from ol_openedx_course_export.constants import AWS_S3_DEFAULT_URL_PREFIX
+
+
+def is_bucket_configuration_valid():
+    """
+    For course export to work properly we need all the AWS settings configured properly
+    """
+    return (
+        settings.COURSE_IMPORT_EXPORT_BUCKET is not None
+        and settings.COURSE_IMPORT_EXPORT_BUCKET.strip() != ""
+    )
+
+
+def get_file_name_with_extension(course_id):
+    """
+    Args:
+        course_id: (str) course_id of the generated course tarball/OLX file
+        e.g. 'course-v1:edX+DemoX+Demo_Course'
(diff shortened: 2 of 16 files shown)


src/ol_openedx_course_export/utils.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def is_bucket_configuration_valid():
+    """
+    For course export to work properly we need all the AWS settings configured properly
+    """
+    return None not in [
+        settings.COURSE_IMPORT_EXPORT_BUCKET,
+        settings.AWS_ACCESS_KEY_ID,
+        settings.AWS_SECRET_ACCESS_KEY,
```
blarghmatey, line 13: We should allow for empty AWS keys so that we can use IAM instance role permissions on the EC2 machines.

src/ol_openedx_course_export/utils.py
```diff
@@ -0,0 +1,30 @@
...
+def get_clean_file_name(tarball_name):
+    """
+    Args:
+        tarball_name: (str) name of the generated course tarball/OLX file
+        e.g. '/tmp/course.l8dtvuog.tar.gz'
+    Returns:
+        str: Returns the file name without prefix
+    """
+    # Since the generated tarball has associated prefix path e.g. "/tmp/" so we trim the path and pick the name
+    return tarball_name.rsplit("/")[-1]
```
blarghmatey, line 26: Would it be possible to build a path name that incorporates the course ID? If it's more than a ~15 minute effort then we'll punt on it for now.

arslanashraf7: @blarghmatey I am a little confused on this but let me try to see the possibility of what you are thinking to make sure I'm in the right direction?

1) Are you referring to `/tmp/course.l8dtvuog.tar.gz` to incorporate course ID e.g. you want the tarball name to be something like `/tmp/course-v1:edX+DemoX+Demo_Course.tar.gz` instead of `/tmp/course.l8dtvuog.tar.gz`?

* This might require changes within the platform itself because the implementation doesn't support the configurable path generation. You can see the implementation [here](https://github.com/openedx/edx-platform/blob/open-release%2Fmaple.master/cms/djangoapps/contentstore/tasks.py#L338-L340). This change would require us to create our own tarball generation implementation though. (More than ~15 min)

OR 

2) If our end goal is to upload the files to s3 with Course ID as the name instead of the generated tarball's name. 
* We can customize this in the plugin and that should not be more than 15 mins. With this change the files uploaded on s3 will have course Id as name and the return URL in response would look like `"course-v1:MIT+MIT_IK_0001+R1": "https://defaultcourseexports.s3.amazonaws.com/course-v1:edX+DemoX+Demo_Course.tar.gz.tar.gz"` (Though we might have to omit some characters e.g. `-` that would not support a URL pattern)

src/ol_openedx_course_export/utils.py
arslanashraf7: I will remove these from the configuration checks except `settings.COURSE_IMPORT_EXPORT_BUCKET`. moreover, even if there would be an error in access it would be logged within the API exception catch.

src/ol_openedx_course_export/utils.py
blarghmatey: Thanks for the clarification, I don't care what the file name is on disk as long as we can customize how it gets represented once it lands in S3.

arslanashraf7: This has been incorporated. now the uploaded files on s3 would have course_id as their name.

src/ol_openedx_course_export/utils.py
arslanashraf7: This is incorporated.

arslanashraf7:
@blarghmatey This is ready for your review.

src/ol_openedx_course_export/LICENSE.txt
```diff
@@ -0,0 +1,672 @@
+
+                    GNU AFFERO GENERAL PUBLIC LICENSE
```
blarghmatey, line 2: We typically use the BSD-3-Clause license as it is compatible with the GPL/AGPL without being as restrictive. That should already be present in the root of the repository.

arslanashraf7: Updated, I've also changed the copyright to use 2022 instead of 2021.

In addition, our other plugins like `ol-openedx-canvas-integration` and `ol-openedx-rapid-response-reports` are also using GNU licenses which I think should be updated in another PR along with copyright 2022 in the root's `license.txt`.

blarghmatey:
:+1:

Issue #28, opened 2022-02-10
Create a plugin that maps SKUs to course ids...


pdpinch:
and returns course ids in ecommerce URLs instead of SKUs. 

see this discussion: https://github.com/mitodl/mitxonline/issues/340#issuecomment-1035214336

### some research questions to start:  

- [ ] How are SKU's set normally on open edX?
- [ ] Assuming that's there's a table SKUs, should the plugin create a SKU every time a course is created?
- [ ] Or should we override an existing API so that the course id is returned instead of looking up the SKU?

arslanashraf7:
Update:

The initial details regarding SKU question in the ticket description are:

>How are SKU's set normally on open edX?

The SKU is simply a `Char` field in the `CourseMode` table and is not really referred to as a relation to some other table [Model Reference](https://edx.readthedocs.io/projects/open-edx-building-and-running-a-course/en/named-release-cypress/running_course/manage_course_fees.html#add-course-mode).

_This leads to a non-plugin solution as well_, Which means we can set our `Product_Id` from the MITxOnline as a value for the SKU field to the relevant course mode in the Open edX. And based on [this comment](https://github.com/mitodl/mitxonline/issues/340#issuecomment-1033683766). When a URL like `https://<marketing_site_base>/checkout/?sku=8CF08E5` is hit, we know that the SKU here is actually the productId from our marketing site. But this will require us to map the product within then edx while adding CourseMode entries.

> Assuming that's there's a table SKUs, should the plugin create a SKU every time a course is created?

SKUs are not part of an independent table they are part of [CourseModes table](https://github.com/openedx/edx-platform/blob/open-release/maple.master/common/djangoapps/course_modes/models.py#L42). As for how they are being set is related to how we generally set the price. It looks like they are added manually while adding a Course Mode through django admin. [Documentation](https://edx.readthedocs.io/projects/open-edx-building-and-running-a-course/en/named-release-cypress/running_course/manage_course_fees.html#add-course-mode)

But I also like @rhysyngsun 's approach as well wherein the plugin will generate a URL for the marketing site with added course id. e.g. On the redirect from plugin, the generated URL will take the user to something like `https://<marketing_site_base>/checkout/?course_id=course-v1-...`. 
And in the marketing site, we decide to fetch the product id relevant to course_id and perform checkout stuff.

@rhysyngsun Is my above understanding correct?

FYI,  @pdpinch

rhysyngsun:
[...]

PR #31, opened 2022-02-21
feat: add external checkout api plugin

arslanashraf7:
#### Pre-Flight checklist

- [x] Testing
  - [ ] Code is tested
- [ ] Settings
  - [ ] New settings are documented and present in `app.json`
  - [ ] New settings have reasonable development defaults, if applicable
  - [ ] Opened issue for DevOps regarding necessary configuration changes to deployed environments

#28 

#### What's this PR do?
- Adds an API to map SKU to external checkout pages.
- Adds a new configuration to set custom external checkout URL for the marketing site
Based on the above, The API works as the alternate to the built-in eCommerce of edX by taking the upgrade URLs request in place of edX's eCommerce, customizes the URL w.r.t the external website with added params, and redirecting the user to that.

You can test the API with steps mentioned in the README (https://github.com/mitodl/open-edx-plugins/tree/arslan/28-external-checkout/src/ol_openedx_checkout_external)

_for TLDR;_
1) Generate & install the plugin with [common instructions](https://github.com/mitodl/open-edx-plugins/tree/arslan/28-external-checkout/src/ol_openedx_checkout_external#installation)
 2) In `private.py` set `MARKETING_SITE_CHECKOUT_URL=http://mitxonline.odl.local:8013/checkout/` # This setting is not part of Open Edx, It's added by this plugin
3) In `private.py` set `ECOMMERCE_PUBLIC_URL_ROOT = 'http://localhost:18000'` (Because we want to use external eCommerce using this API plugin for redirection)
4) Create an new ecommerce configuration in `http://<LMS_BASE>/admin/commerce/commerceconfiguration` with "basket_checkout_page=/checkout-external/"
5) Create a CourseMode entry for your test course by visiting `http://localhost:18000/admin/course_modes/coursemode/` and adding some value for the SKU.
6) Enroll in your test course and click `Upgrade` button in the dashboard or learning MFE. You should be redirected to the checkout page of MITxOnline e.g. (`http://mitxonline.odl.local:8013/checkout/?course_id=test_course_id`).

Plugin build generation and configurations

### DIFF
--- a/src/ol_openedx_checkout_external/views.py
+++ b/src/ol_openedx_checkout_external/views.py
@@ -1,0 +1,67 @@
+"""Views for External Checkout"""
+
+import logging
+
+from common.djangoapps.course_modes.models import CourseMode
+from django.conf import settings
+from django.contrib.auth.decorators import login_required
+from django.http import Http404, HttpResponseRedirect
+
+from ol_openedx_checkout_external.exceptions import ExternalCheckoutError
+
+log = logging.getLogger(__name__)
+
+
+@login_required
+def external_checkout(request):
+    """
+    An API View for external checkout that redirects to the marketing site for checkout
+
+    **Example Requests**
+
+    GET /checkout-external/?sku=TESTSKU
+
+    **Example Responses**
+
+    The API will return two types of response codes in general
+    302, 404, 500
+
+    A 302 redirect for the marketing site checkout would be returned in case the request was successful
+
+    A 404 would be returned in case there is no sku matching products
+
+    A 500 would be returned if there are any configuration errors
+    """
+
+    if request.method != "GET":
+        raise NotImplementedError("API only supports GET requests")
+
+    if not settings.MARKETING_SITE_CHECKOUT_URL:
+        raise ExternalCheckoutError(
+            "MARKETING_SITE_CHECKOUT_URL value is not configured properly"
+        )
+
+    product_sku = request.GET.get("sku")
+
+    if not product_sku:
+        log.error("No Product SKU was found")
+        raise Http404
+
+    course_modes = CourseMode.objects.filter(sku=product_sku)
+    if not course_modes:
+        log.error(
+            f"No CourseMode was found against the given product SKU ({product_sku})"
+        )
+        raise Http404
+
+    # Because there is no unique constraint on SKU, so there could be multiple CourseModes with same SKU
+    if len(course_modes) > 1:
+        raise ExternalCheckoutError(
+            f"Found multiple CourseModes for the same SKU ({product_sku})"
+        )
+
+    #  Generate a URL to redirect to marketing site based on its checkout URL with and added
+    #  course ID query param)
+    course_id = str(course_modes.first().course.id)
+    redirect_url = f"{settings.MARKETING_SITE_CHECKOUT_URL}?course_id={course_id}"
+    return HttpResponseRedirect(redirect_to=redirect_url)
--- a/src/ol_openedx_checkout_external/.coveragerc
+++ b/src/ol_openedx_checkout_external/.coveragerc
@@ -1,0 +1,10 @@
+[run]
+branch = True
+data_file = .coverage
+source=ol_openedx_checkout_external
+omit =
+    test_settings
+    *migrations*
+    *admin.py
+    *static*
+    *templates*
--- a/src/ol_openedx_checkout_external/BUILD
+++ b/src/ol_openedx_checkout_external/BUILD
@@ -1,0 +1,18 @@
+python_sources(
+    name="external_checkout",
+    dependencies=["src/ol_openedx_checkout_external/settings:checkout_external_settings"]
+)
+
+python_distribution(
+    name="checkout_external_package",
+    dependencies=[":external_checkout"],
+    provides=setup_py(
+        name="ol-openedx-checkout-external",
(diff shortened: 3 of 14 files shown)


src/ol_openedx_checkout_external/views.py
```diff
@@ -0,0 +1,96 @@
...
+from openedx.core.lib.api.authentication import BearerAuthenticationAllowInactiveUser
+from openedx.core.lib.api.view_utils import DeveloperErrorViewMixin
+from rest_framework import status
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.views import APIView
+
+log = logging.getLogger(__name__)
+
+
+class ExternalCheckoutView(DeveloperErrorViewMixin, APIView):
```
rhysyngsun, line 21: I don't think this should be an APIView, because if something goes wrong the user is going to just get JSON back in their browser rather than a html styled error page. I think this changes some of the response types later on, which I'll make note of what I'm thinking those should be.

src/ol_openedx_checkout_external/views.py
```diff
@@ -0,0 +1,96 @@
...
+    )
+    permission_classes = (IsAuthenticated,)
+    http_method_names = ["get"]
+
+    def get(self, request):
+        """
+        External checkout request
+        """
+        if settings.MARKETING_SITE_CHECKOUT_URL is None or "":
+            raise self.api_error(
```
rhysyngsun, line 52: This should be a 500 error, because it's a configuration error.

src/ol_openedx_checkout_external/views.py
```diff
@@ -0,0 +1,96 @@
...
+        if not course_modes:
+            raise self.api_error(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                developer_message=f"CourseMode doesn't exist for SKU ({product_sku})",
+                error_code="internal_error",
+            )
+
+        # Because there is no unique constraint on SKU, so there could be multiple CourseModes with same SKU
+        if len(course_modes) > 1:
+            raise self.api_error(
```
rhysyngsun, line 77: This should be a 500 error because it's a configuration error.

src/ol_openedx_checkout_external/views.py
```diff
@@ -0,0 +1,96 @@
...
+        try:
+            #  Generate a URL to redirect to marketing site based on its checkout URL with and added
+            #  course ID query param)
+            course_id = str(course_modes.first().course.id)
+            redirect_url = (
+                f"{settings.MARKETING_SITE_CHECKOUT_URL}?course_id={course_id}"
+            )
+            return HttpResponseRedirect(redirect_to=redirect_url)
+        except Exception as e:
+            raise self.api_error(
```
rhysyngsun, line 92: No need to catch this, whatever error happens can just bubble up into a 500.

src/ol_openedx_checkout_external/views.py
```diff
@@ -0,0 +1,96 @@
...
+        SessionAuthenticationAllowInactiveUser,
+    )
+    permission_classes = (IsAuthenticated,)
+    http_method_names = ["get"]
+
+    def get(self, request):
+        """
+        External checkout request
+        """
+        if settings.MARKETING_SITE_CHECKOUT_URL is None or "":
```
rhysyngsun, line 51: Did you mean to have this raise this error if `MARKETING_SITE_CHECKOUT_URL` is either `None` or `""`? Python is evaluating this as `(settings.MARKETING_SITE_CHECKOUT_URL is None) or ""`, so if  `MARKETING_SITE_CHECKOUT_URL=""`, then that turns into `("" is None) or ""` -> `False or ""` -> `""`, which is falsy.

I think `if not settings.MARKETING_SITE_CHECKOUT_URL` is what you're looking for here.

rhysyngsun (changes requested):

arslanashraf7:
@rhysyngsun The feedback has been addressed. You were right the users won't need JSON in this API.

Since almost all of the errors in the API would not be usable for the end-user because he would have no idea about how the API will behave because all the settings are needed to be done by admins/SRE. So, I've raised exceptions in those cases to show 500s template error pages to the end-users and an exception message for the SRE/Developers to see what went wrong. 

This PR is ready for another review, Let me know your thoughts.

rhysyngsun (approved):
I've been having difficulties with my local openedx install, but i don't want to hold you up on this anymore so I'm going to say it LGTM on a code review basis.

arslanashraf7:
>I've been having difficulties with my local openedx install, but i don't want to hold you up on this anymore so I'm going to say it LGTM on a code review basis.

@rhysyngsun Thanks for reviewing the PR. We might be able to install test it on the QA instance.

> Can you fix the failing precommit checks though?

Yes, I remember pre-commit complaining about secrets being shared some time ago but looking around I noticed that it has been fixed and I thought the only thing this PR might need will be rebased. 

But after the rebase the pre-commit just broke. It's not even getting past the build step which seems to be an issue with the image that the pre-commit pulling to use & runs its checks on. I couldn't see more details on the pre-commit dashboard due to access. @blarghmatey Do you have any thoughts on this?

blarghmatey:
Pre-commit is passing again now. I updated the version on the main branch and then rebased here.

arslanashraf7:
Thanks for making the change Tobias. I also checked the package build etc. Works fine.

PR #33, opened 2022-03-14
perf: enhance course export to use celery tasks

arslanashraf7:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/open-edx-plugins/issues/27

#### What's this PR do?
- As the next part of the course export enhancement we attempted running the course export in separate celery tasks to avoid any timeout issues that we see. (This is something similar that the existing course export functionality is doing, and we intend to follow the same track in our plugin)
- Trying to implement course export tasks and structure that would match Open edX's local course export but support the celery tasks for exporting courses to s3.
- Add a new API to `GET` the status of the running tasks by course key & task ID

- Install the plugin through the steps mentioned in the Readme of the plugin (Or the steps mentioned in https://github.com/mitodl/open-edx-plugins/pull/29)
- Configure the required AWS keys mentioned in https://github.com/mitodl/open-edx-plugins/pull/29

**1) Now test the POST API with an admin User** 
    - API: `http://localhost:18010/api/courses/v0/export/`
    -  Sample Payload: `{"courses": ["course-v1:MIT+MIT_IK_0001+R1", "course-v1:MIT+MIT_RR_002+R1"}`
    - Sample Response: 
```
{
    "upload_urls": {
        "course-v1:MIT+MIT_IK_0001+R1": "https://defaultcourseexports.s3.amazonaws.com/course-v1:MIT+MIT_IK_0001+R1.tar.gz",
        "course-v1:MIT+MIT_RR_002+R1": "https://defaultcourseexports.s3.amazonaws.com/course-v1:MIT+MIT_RR_002+R1.tar.gz",
    },
    "upload_task_ids": {
        "course-v1:MIT+MIT_IK_0001+R1": "0bc21ee1-1679-407f-839d-3527f38535d6",
        "course-v1:MIT+MIT_RR_002+R1": "dfbc05c1-30f6-4f23-b7f7-d101079cf27d",
    },
    "failed_uploads": {}
}
```

**2) Now test the new GET API:**
    - API: `http://localhost:18010/api/courses/v0/export/<course_id>/?task_id=<task_id>` (You will get both in the result of above POST request)
    - Test the response which could be something like 
```
{
    "state": "Succeeded"
}
(Other responses could be `Failed`, `In Progress`)
``` 

- Plugin installation and configurations

We recently implemented course export to s3 feat. The feat was initially implemented as a `synchronous operation` because it was going to run within a Pipeline and keeping it synchronous and returning the upload results when the API was successful was an initial requirement. But we saw some timeout issues when the course sizes are big. So, we decided to enhance the plugin to use celery tasks to upload the course. More details on https://github.com/mitodl/open-edx-plugins/issues/27 & https://github.com/mitodl/open-edx-plugins/pull/29

**Steps to setup celery worker locally:**
1. Set this if it is not already set in devstack_with_worker.py
`BROKER_URL = 'redis://:password@edx.devstack.redis:6379/'`
2. update docker-compose file of devstack and change cms service setting to devstack_with_worker
3. `make dev.up.redis`
4. dev.stop
5. dev.up
6. in new terminal run celery worker using this command `DJANGO_SETTINGS_MODULE=cms.envs.devstack_with_worker celery worker --app=cms.celery:APP -l=DEBUG`

It appears edx has wrong values for  CELERY_QUEUES  setting in /edx/etc/lms.yml
 if we change the value of that setting to this workers work as expected
  'edx.lms.core.default',
    'edx.lms.core.high',
    'edx.lms.core.high_mem'

might have to set C_FORCE_ROOT environment variable to allow running celery worker with root user

### Reviewer Note:
1) After this PR An email would be sent upon task completion for every course, this will happen because we are reusing some part of the existing course export from Open Edx. For 

**Some References:**
- Email sending implementation task that runs after a user task completion single is sent [Ref](https://github.com/openedx/edx-platform/blob/master/cms/djangoapps/cms_user_tasks/tasks.py#L24)
- Existing course export celery based implementation can be seen in [Here](https://github.com/openedx/edx-platform/blob/master/cms/djangoapps/contentstore/views/import_export.py#L337) and [Here](https://github.com/openedx/edx-platform/blob/master/cms/djangoapps/contentstore/tasks.py#L294)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -39,7 +39,7 @@ repos:
     - --remove-unused-variables
     - --remove-all-unused-imports
 - repo: https://github.com/Yelp/detect-secrets
-  rev: v1.1.0
+  rev: v1.2.0
   hooks:
   - id: detect-secrets
     args:
--- a/src/ol_openedx_course_export/s3_client.py
+++ b/src/ol_openedx_course_export/s3_client.py
@@ -5,11 +5,13 @@ from ol_openedx_course_export.utils import get_file_name_with_extension
 
 
 class S3Client:
-    def __init__(self):
-        self.client = self.get_s3_client()
+    client = None
 
-    @staticmethod
-    def get_s3_client():
+    def __init__(self):
+        if not self.client:
+            self.client = self.get_s3_client()
+
+    def get_s3_client(self):
         return boto3.resource(
             "s3",
             aws_access_key_id=settings.AWS_ACCESS_KEY_ID,
--- a/src/ol_openedx_course_export/tasks.py
+++ b/src/ol_openedx_course_export/tasks.py
@@ -1,0 +1,49 @@
+"""
+This file contains celery tasks related to S3 course export plugin.
+"""
+import logging
+
+from botocore.exceptions import ClientError
+from celery import shared_task  # pylint: disable=import-error
+from cms.djangoapps.contentstore.tasks import CourseExportTask, create_export_tarball
+from opaque_keys.edx.keys import CourseKey
+from user_tasks.models import UserTaskStatus
+from xmodule.modulestore.django import modulestore
+
+from ol_openedx_course_export.s3_client import S3Client
+
+log = logging.getLogger(__name__)
+
+
+@shared_task(base=CourseExportTask, bind=True)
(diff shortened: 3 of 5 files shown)


blarghmatey:
Looks good to me :+1:

PR #49, opened 2022-04-19
docs: update BSD licenses

arslanashraf7:
None - I just noticed that the licenses in some older plugins were not BSD-based (Specifically BSD-3-Clause). 

#### What's this PR do?
- Updates the packages to have BSD-3 licenses
- Quick typo fix

None, Just a general look around the PR

Just check the documentation

- We added some packages earlier where we didn't add the BSD-3 clause licenses instead we used GNU based license.


blarghmatey (approved):
:+1: Thanks for catching that

Issue #13, opened 2021-11-15
Canvas Integration: Serve HTML/JS/CSS from within the plugin


arslanashraf7:
**Related to:**
https://github.com/mitodl/open-edx-plugins/issues/7#issuecomment-956173887 & https://github.com/mitodl/open-edx-plugins/issues/7#issuecomment-956292339

**Description:**
We have majorly migrated the canvas integration code from platform to plugin in https://github.com/mitodl/open-edx-plugins/pull/8. Since the instructor dashboard doesn't currently support as mentioned [here](https://github.com/mitodl/open-edx-plugins/pull/8). As a next step, we need to give a try to https://github.com/edx/web-fragments and see if we could pass through the html/js/css from within the plugin using the fragments(https://github.com/edx/web-fragments) and plugin contexts (https://github.com/edx/edx-django-utils/blob/master/edx_django_utils/plugins/README.rst).

**Experimantal Progress**
On the way of doing some experimental work we were able to migrate [some part](https://github.com/mitodl/edx-platform/commit/b61152ac70#diff-0d1ee66fb84bee6c3517980a7cefd74a050ccb60da9c0c3b662ab0a2d80d17c0R691) of the instructor dashboard from platform to the plugin and created [this branch in platform](https://github.com/mitodl/edx-platform/tree/arslan/canvas-context) and [this branch in open-edx-plugins](https://github.com/mitodl/open-edx-plugins/tree/arslan/canvas-plugin-context)

pdpinch:
@arslanashraf7 is there anything left to do on this issue?

arslanashraf7:
This is complete. Thanks for checking back on this.

PR #54, opened 2022-05-16
feat: serve assets from plugin - nutmeg support

arslanashraf7:
#### Pre-Flight checklist

- [x] Screenshots and design review for any changes that affect layout or styling
  - [x] Desktop screenshots
- [ ] Migrations
  - [ ] Migration is backwards-compatible with current production code
- [ ] Testing
  - [ ] Code is tested

#13 

#### What's this PR do?
- Integrations the context api
- Adds the assets within the plugin
- adds configurations to serve the assets from the plugin
- will add configurations for pants build

- Setup edx-platform on nutmeg release
- cherry-pick the commit (A shortened version of the commit will be added once the PR is finalized because the current cherry-pick mentioned in this plugin's readme has all the things that won't be needed after this migration but some will still be there)
- Install the plugin

- Setting up edx-platform on Nutmeg

#### RELEASE NOTE:
- We might need to compile assets(`make lms-static`) once this plugin is released and installed on the platform. Otherwise we might notice assets issues like we used to face in past.

When we added this plugin in the first place, edX won't support serving the assets right from within the plugin. That support was added in the recent Nutmeg release so it's the right time to migrate this plugin and reduce it's cherry-picking to minimum

#### Screenshots (if appropriate)
Screenshot 2022-05-18 at 1 44 54 PM

### DIFF
--- a/src/ol_openedx_canvas_integration/BUILD
+++ b/src/ol_openedx_canvas_integration/BUILD
@@ -3,12 +3,24 @@ python_sources(
     dependencies=["src/ol_openedx_canvas_integration/settings:canvas_settings"],
 )
 
+resources(
+  name="canvas_plugin_templates",
+  sources=[],
+  dependencies=["src/ol_openedx_canvas_integration/templates:canvas_templates"]
+)
+
+resources(
+  name="canvas_plugin_js",
+  sources=[],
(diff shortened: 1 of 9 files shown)


blarghmatey:
This looks good to me :+1: 

I ran the pants package and verified that the JS and HTML assets are included.

Issue #110, opened 2023-03-23
Add REST API for course structure JSON


blarghmatey:
# User Story
- As a data engineer I want to be able to retrieve the JSON representation of course structure so that I can include that in a nightly export of edX data

# Description/Context
There is a management command defined [here](https://github.com/openedx/edx-platform/blob/master/lms/djangoapps/courseware/management/commands/dump_course_structure.py) that will generate a JSON export of the structure for a given course. Rather than requiring a shell process in a running edX instance I would like to be able to retrieve that information as a REST API that accepts a course ID as a parameter. 

@pdpinch should the structure of a course be considered "sensitive" information that would require authentication? I would assume yes.

# Acceptance Criteria
- [ ] The JSON structure of a course can be retrieved as a REST API call with a course ID as the parameter
- [ ] Optional: A bulk JSON course structure API for all courses in a given org

pdpinch:
Instead of exporting the data structure nightly, what do you think about exporting it every time the course is published? This would cut down on duplicate copies in the export, and duplication of effort in the ETL.

blarghmatey:
I'm not opposed to that approach, but I'm unclear on what you mean by "duplicate copies in the export". This document is not included as part of the course OLX export if that's what you're referring to. As to duplication of effort in the ETL, this is not currently part of the nightly data extracts.

I'm also unclear on the mechanics of how the push-based export on publish would be integrated into the data exports. It's certainly not impossible, I just don't know off-hand how that would function. Keeping it as a pull-based request in a REST API is simpler from a pipeline design standpoint.

I'm happy to discuss further.

rachellougee:
Re-open it as I need to fix these ImportError in https://mit-office-of-digital-learning.sentry.io/issues/?project=1730882&referrer=sidebar.

This [PR](https://github.com/openedx/edx-platform/commit/9d8375ff99f999e4c09220d85df368c78bb14bce#diff-a61cdcef8d334dae9af48ecc9a38e2157c2683ad23bc2d34806528244a9430d9R81) had a function name change (`dump_module ` to `dump_block`) that this plugin depends on. It was released in https://github.com/openedx/edx-platform/releases/tag/release-2023-03-08-14.35.  But `xpro` still uses `olive` which still use the old name [dump_module](https://github.com/openedx/edx-platform/blob/open-release/olive.master/lms/djangoapps/courseware/management/commands/dump_course_structure.py#L81)

PR #203, opened 2023-10-04
fix ImportError on dump_block for olive release 

rachellougee:

https://github.com/mitodl/open-edx-plugins/issues/110

This PR fixes ImportError for `olive` open edx instance (e.g. xpro) https://mit-office-of-digital-learning.sentry.io/issues/?project=1730882&referrer=sidebar
```
cannot import name 'dump_block' from 'lms.djangoapps.courseware.management.commands.dump_course_structure' (/openedx/edx-platform/./lms/djangoapps/courseware/management/commands/dump_course_structure.py
```

In older versions of open edx like `olive` it still uses the old function name `dump_module` in [this file](https://github.com/openedx/edx-platform/blob/open-release/olive.master/lms/djangoapps/courseware/management/commands/dump_course_structure.py#L81). But `dump_module` has renamed to `dump_block` in [release-2023-03-08-14.35](https://github.com/openedx/edx-platform/releases/tag/release-2023-03-08-14.35). For backward compatibility, this PR handles ImportError to import either `dump_module` or `dump_block` 

# How can this be tested?

Follow the step in [README](https://github.com/mitodl/open-edx-plugins/tree/main/src/ol_openedx_course_structure_api#installation) 

then I modified Dockerfile in tutor/env/build/openedx/ to point to https://github.com/openedx/edx-platform/tree/open-release/olive.master.
```
FROM minimal as code
ARG EDX_PLATFORM_REPOSITORY=https://github.com/openedx/edx-platform.git
ARG EDX_PLATFORM_VERSION=open-release/olive.master
```
Ran `tutor images build openedx`
Ran `tutor local run lms ./manage.py lms print_setting INSTALLED_APPS` to make sure no ImportError

To test palm, I changed Dockerfile back to what was before (latest release)
```
FROM minimal as code
ARG EDX_PLATFORM_REPOSITORY=https://github.com/openedx/edx-platform.git
ARG EDX_PLATFORM_VERSION=open-release/palm.2
```
Ran the same steps to make sure no error

# Additional Context

### DIFF
--- a/src/ol_openedx_course_structure_api/views.py
+++ b/src/ol_openedx_course_structure_api/views.py
@@ -2,9 +2,6 @@
 
 import logging
 
-from lms.djangoapps.courseware.management.commands.dump_course_structure import (
-    dump_block,
-)
 from opaque_keys import InvalidKeyError
 from opaque_keys.edx.keys import CourseKey
 from openedx.core.lib.api.view_utils import (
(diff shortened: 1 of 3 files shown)


src/ol_openedx_course_structure_api/views.py
```diff
@@ -86,6 +83,21 @@ def get(self, request, course_id):
         if inherited_metadata:
             compute_inherited_metadata(course)
 
+        # dump_block is renamed from dump_module in release-2023-03-08-14.35
+        # For backward compatibility, try import the new name and then old name
+        try:
```
blarghmatey, line 88: Should this be done in the top level imports, rather than in the context of the function execution?

rachellougee: I did it inside right before `dump_block` is called. Probably can be done in top level as well. I can test. Is it preferred to be done in top level?

rachellougee: Moved to the top level and tested

blarghmatey (approved):

Issue #230, opened 2023-11-08
Upgrade to support Django 4.x


arslanashraf7:
# Description/Context
The Quince release of edX has migrated to the 4.2 release of Django. The plugins in our mono repo still use older classes/functions from Django e.g. deprecated `urls` etc.

# Plan/Design
-  Check and fix if a plugin could downgrade django to 3.2 
- Provide any code modifications necessary to work with the newer Django release.

arslanashraf7:
Done via #231

PR #231, opened 2023-11-08
chore: support django 4.x

arslanashraf7:
https://github.com/mitodl/open-edx-plugins/issues/230

- Removes deprecated/removed functionality from old Django 3.x
- Support for 4.x

# How can this be tested?
-  Package geenration should be successful
- Everything should work with Quince release of edX

### DIFF
--- a/src/edx_sysadmin/BUILD
+++ b/src/edx_sysadmin/BUILD
@@ -21,7 +21,7 @@ python_distribution(
     ],
     provides=setup_py(
         name="edx-sysadmin",
-        version="0.2.3",
+        version="0.3.0",
         description="An Open edX plugin to enable SysAdmin panel",
         license="BSD-3-Clause",
         author="MIT Office of Digital Learning",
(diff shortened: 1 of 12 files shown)


blarghmatey (commented):
This all looks good. Can you also bump the versions in the `BUILD` files for the affected packages?

arslanashraf7:
> This all looks good. Can you also bump the versions in the `BUILD` files for the affected packages?

Sure, Let me do that in the next commit

arslanashraf7:
@blarghmatey Bumped the versions in https://github.com/mitodl/open-edx-plugins/pull/231/commits/89e9556ef479f5b7d119d7ea89aa858b140485c4. I intentionally changed the minor ones instead of patch because this is going as a support for major Django version. What's your opinion?

blarghmatey:
I think that makes sense. That's what I did for rapid-response as well.

blarghmatey (approved):

PR #262, opened 2024-01-05
Adding README.md

Ardiea:
https://github.com/mitodl/ol-infrastructure/issues/1976

Added a README detailing dangers of not incrementing version numbers + linking to the build and publish pipeline.

# How can this be tested?
N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,7 @@
+# Important
+
+Ensure you update the version number in `src/<plugin>/BUILD` before any merge to `main`. Failure to do so will overwrite the existing, published version. 
+
+# Build and Publish
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -0,0 +1,7 @@
+# Important
+
+Ensure you update the version number in `src/<plugin>/BUILD` before any merge to `main`. Failure to do so will overwrite the existing, published version.
```
blarghmatey, line 3: PyPI actually won't let you overwrite a published package. What would happen is that the pipeline `put` step would fail due to the version conflict.

arslanashraf7:
Just came across this PR. Thanks for adding these details @Ardiea . We definitely needed these types of details in our docs/readme.

PR #282, opened 2024-01-30
fix: update Otel plugin dependencies

shahbaz-shabbir05:
The problem is that when installing the OpenTelemetry (Otel) Plugin, some necessary dependencies related to OpenTelemetry are not automatically installed. Even though these dependencies are listed in the `pyproject.toml` file, which is the same for all plugins in this repo, they still don't get installed as expected. Additionally, the `setup.py` file, which is for plugin setup, lacks any mention of OpenTelemetry packages under install_requires.

To solve this issue:

- The required packages are now explicitly mentioned in a `BUILD` file under dependencies for the Otel Plugin. This ensures that the necessary packages are included during the plugin setup.

- There have been updates to some packages related to OpenTelemetry. Instead of individually installing the `API` and `SDK` packages, a new approach involves using the `OpenTelemetry Distro` package. This Distro package includes all the required components, simplifying the installation process.


blarghmatey:
Thanks for tracking this down and for the fix!

blarghmatey (approved):

PR #328, opened 2024-05-15
feat: added mitxpro-openedx-extensions

Anas12091101:
Closes https://github.com/mitodl/hq/issues/4142

Adds mitxpro-openedx-extensions in open-edx-plugins

### How can this be tested?
- Start edx-platform and mitxpro containers.
- Clone this project in the src folder present in your Openedx work directory.
- Install pants locally and run the command in the project folder: `pants package ::`. This will create a tar file `openedx-companion-auth-1.3.0.tar.gz` in the dist folder.
- Configure openedx with mitxpro using https://github.com/mitodl/mitxpro/blob/master/docs/configure_open_edx.md
- Install the tar file in your lms by first creating the lms shell using `make lms-shell` command in the devstack folder and then `pip install /edx/src/open-edx-plugins/dist/openedx-companion-auth-1.3.0.tar.gz`.
- Open a private window and go to localhost:18000. You should be redirected to mitxpro signin page.

Note: Tested this on both master and quince release


run_devstack_integration_tests.sh
Anas12091101, line 1: Changed some test logic here. Previously all the plugins were being installed at once in an outer for loop. Due to this, the tests of edx-sysadmin started failing as they required login and edx-platform keep redirecting them to mitxpro signin page. Now, the plugins are installed only when we run their tests and are uninstalled after the tests run.

src/openedx_companion_auth/README.rst
arslanashraf7, line 1: The readme can be updated to include the features like other plugins. I think the README is copied from top level repo?

src/openedx_companion_auth/apps.py
```diff
@@ -0,0 +1,22 @@
...
+    verbose_name = "Openedx Companion Auth"
+
+    plugin_app = {
+        "settings_config": {
+            "lms.djangoapp": {
+                "test": {"relative_path": "settings.test"},
+                "common": {"relative_path": "settings.common"},
+                # aws deprecated in favor of production, see https://openedx.atlassian.net/browse/DEPR-14
+                "aws": {"relative_path": "settings.production"},
+                "production": {"relative_path": "settings.production"},
```
arslanashraf7, line 19: I think we need just the common and production keys here.

src/openedx_companion_auth/manage.py
arslanashraf7, line 1: We don't need this file.

src/openedx_companion_auth/settings/common.py
```diff
@@ -0,0 +1,24 @@
+"""
+Settings for openedx-companion-auth
+"""
+
+SECRET_KEY = "unsafe-secret-key"  # noqa: S105  # pragma: allowlist secret
+INSTALLED_APPS = (
+    "django.contrib.auth",
+    "django.contrib.contenttypes",
+    "django.contrib.sessions",
+)
```
arslanashraf7, line 10: I don't think we need these?

arslanashraf7 (changes requested):
This is code review, I'm going to test the plugin now.

src/openedx_companion_auth/BUILD
```diff
@@ -0,0 +1,30 @@
...
+)
+
+python_distribution(
+    name="openedx_companion_auth_package",
+    dependencies=[
+        ":openedx_companion_configuration",
+    ],
+    provides=setup_py(
+        name="openedx-companion-auth",
+        version="1.3.0",
```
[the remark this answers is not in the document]
Anas12091101: I think @blarghmatey could give a better answer to this

run_devstack_integration_tests.sh
```diff
@@ -14,34 +14,70 @@ paver update_assets lms --settings=test_static_optimized
...
+            # Check for the existence of settings/test.py
+            if [ -f "settings/test.py" ]; then
+                pytest_command="pytest . --cov . --ds=settings.test"
+            else
+                pytest_command="pytest . --cov ."
+            fi
+
+            # Run the pytest command
+            $pytest_command
+
```
arslanashraf7, line 80: I think we can make this one simpler, Without `manage_plugin` function because we don't need to uninstall the package after running the tests. You can let me know if there is a reason we are uninstalling them.

Otherwise what I think we can do in this loop is:

1. Install the plugin, Add the log about which plugin is being tested
2. Run tests
3. Do other things, like coverage etc

src/openedx_companion_auth/README.rst
```diff
@@ -0,0 +1,134 @@
+Open edX auth companion
+=======================
+
+An Openedx auth companion plugin that redirects users from edx-platform to MIT applications
```
arslanashraf7, line 4:
```suggestion
An Openedx plugin that redirects users from edx-platform to MIT applications
```

src/openedx_companion_auth/README.rst
```diff
@@ -0,0 +1,134 @@
...
+
+Building
+--------
+
+- Build a pip-installable package into `dist/`:
+
+  .. code-block:: shell
+
+     pants package ::
+
```
arslanashraf7, line 25: I think we can keep these steps standardized with other plugins, They don't mention the pants installation steps. That should be part of the main Readme.

src/openedx_companion_auth/README.rst
```diff
@@ -0,0 +1,134 @@
...
+Installing
+----------
+
+All that is required to install this in either a hosted version of Open edX or devstack is a `pip install` either from PyPI or the output of `pants build`.
```
arslanashraf7, line 29: The installation steps could also be similar to the other plugins.

src/openedx_companion_auth/README.rst
```diff
@@ -0,0 +1,134 @@
...
+
+  .. code-block:: shell
+
+     pip install /edx/src/open-edx-plugins/dist/openedx-companion-auth-{version}.tar.gz
+
+- Go to the package directory:
+
+  .. code-block:: shell
+
+     cd open-edx-plugins/src/openedx_companion_auth
```
arslanashraf7, line 63: I think all these steps would moved under the Installation heading, You can check the other plugin READMEs for reference. e.g. https://github.com/mitodl/open-edx-plugins/blob/8142a0c654150c21de2602024cc9bb5c2d12fc49/src/edx_sysadmin/README.rst

src/openedx_companion_auth/README.rst
```diff
@@ -0,0 +1,134 @@
...
+      }
+
+  * ``LOCAL_MITX_ALIAS`` should be your ``/etc/hosts`` alias for the mitx application you are configuring this plugin for
+  * ``LOCAL_MITX_PORT`` should be 8053 for MITxPro and 8013 for MITxOnline
+  * ``EXTERNAL_MITX_ONLINE_HOST`` will depend on your OS, but it needs to be resolvable within the edx container
+
+    * Linux users: The gateway IP of the docker-compose networking setup for mitxonline as found via ``docker network inspect mitx-online_default``
+    * OSX users: Use ``host.docker.internal``
+
+  * Save the configuration.
```
arslanashraf7, line 132: This plugin can be tested independently if we properly set the settings e.g. `MITX_REDIRECT_LOGIN_URL `. We don't need to add the whole set of instructions about setting xPRO/edX. One needs to know if the middleware is working as it should. e.g. is it redirecting to the desired URL set in MITX_REDIRECT_LOGIN_URL or not

src/openedx_companion_auth/README.rst
```diff
@@ -0,0 +1,134 @@
...
+
+- Go to ``http://<EDX_HOSTNAME>:18000`` , you should be redirected to the MIT login page
```
arslanashraf7, line 134: Once you address the above comment, this will also need to be changed.

arslanashraf7 (changes requested):

run_devstack_integration_tests.sh
Anas12091101: In the current scenario, yes we don't need to uninstall the plugin after tests because the tests of openedx-companion-auth run at the end after sysadmin and rapid response plugin. But in future if we add a plugin whose tests run after openedx-companion-auth's tests and the plugin requires login to run the test then the openedx-companion-auth will affect that.

src/openedx_companion_auth/BUILD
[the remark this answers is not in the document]
blarghmatey: I think it doesn't hurt to set it to 1.0, but I don't have a strong opinion either way.

run_devstack_integration_tests.sh
arslanashraf7: In a scenario you mentioned above, A plugin that needs `openedx-companion-auth` would have it in it's dependencies. That means the companion package will automatically be installed as part of that plugin.

src/openedx_companion_auth/apps.py
```diff
@@ -0,0 +1,19 @@
+"""openedx_companion_auth app config"""
+
+from django.apps import AppConfig
+
+
+class MITxCoreConfig(AppConfig):
+    """App configuration"""
+
+    name = "openedx_companion_auth"
```
arslanashraf7, line 9: Sorry, I thought I had commented on this earlier. I think a more suitable name w.r.t the other plugins in this mono repo would be `ol_openedx_companion_auth`.

@blarghmatey do you have any reservations about this?

arslanashraf7 (approved):
👍 Thanks for making all the changes. I request some minor feedback. Feel free to squash& merge after that.

Also, We need to do some followup tasks, I'll create a separate ticket(s) for this:

- Update the docs for xPRO and MITx Online
- Deprecate the original repo for this plugin

src/openedx_companion_auth/BUILD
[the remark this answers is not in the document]
arslanashraf7: @rhysyngsun What do you recommend?

Anas12091101:
@blarghmatey openedx-companion-auth is merged and is ready to be moved to pypi.

PR #350, opened 2024-07-01
version: Update Sentry plugin to use SDK v2.x

blarghmatey:
N/A

The latest releases of edX have addressed the infinite recursion bug that was being caused by Sentry, so this updates the Sentry SDK to the latest releases.

This also updates Pants to the latest stable version since it was out of date

### How can this be tested?

Build the latest Sentry plugin and install in a version of edX >= Redwood and ensure that it doesn't cause an infinite recursion in the BulkEmail feature

### DIFF
--- a/src/ol_openedx_sentry/settings/sentry.py
+++ b/src/ol_openedx_sentry/settings/sentry.py
@@ -7,9 +7,6 @@ from functools import partial
 from typing import Any, Optional, Union
 
 import sentry_sdk
-from sentry_sdk.integrations.celery import CeleryIntegration
-from sentry_sdk.integrations.django import DjangoIntegration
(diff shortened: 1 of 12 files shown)


src/ol_openedx_sentry/settings/sentry.py
```diff
@@ -111,11 +108,10 @@ def plugin_settings(app_settings):
         sentry_sdk.init(
             dsn=sentry_dsn,
             environment=env_tokens.get("SENTRY_ENVIRONMENT"),
-            integrations=[DjangoIntegration(), CeleryIntegration(), RedisIntegration()],
```
blarghmatey, line 114: This line is not required with SDK v2 because all of these integrations will be enabled automatically.

shaidar (approved):

PR #393, opened 2024-11-12
feat: add initial architecture for AI chat aside

arslanashraf7:
### What are the relevant tickets?
Related to/Part of/Partial implementation of https://github.com/mitodl/hq/issues/6004

### Description (What does it do?)
- Adds the initial experiments infrastructure for the AI chatbot in open edx

### Screenshots (if appropriate):
NOTE: These screenshots are just to give you and idea, I still need to add some buttons and fields here and there

**Studio test display**

image

**Student/Learner View**
image

### How can this be tested?

-  Install this package in Open edX LMS & CMS
- Enable xBlock configurations in CMS <STUDIO_URL>/admin/xblock_config/studioconfig/
- Enable xBlock asides in LMS <LMS_URL>/admin/lms_xblock/xblockasidesconfig/
- Reload both LMS/CMS
- Go to CMS, Add any mcq problem and you should see a a couple of text edit field button at the bottom
- Got to LMS and you should see a Hello World button

### Additional Context
This is just the basic architecture for the aside we are going to create to add chat support in xBlocks

PR #397, opened 2024-11-19
feat: add basic data fields and design (CMS)

arslanashraf7:
https://github.com/mitodl/hq/issues/6079

- Adds the initial fields required for the chatbot
- Adds the basic form for adding all the configuration fields for the chat:
    - Prompt text
    - Enabled
    - Additional solution
    - GPT model selection (Configurable)

- Adds filtering and allows this aside only on the problem and video blocks
- Enables the user to save the data in these fields

https://github.com/user-attachments/assets/39a90c24-41de-49f6-9511-4e37c3b6b4e4

### How can this be tested?
- Generate the Gzip of this package using pants
- Follow the setup required configurations from https://github.com/mitodl/open-edx-plugins/pull/393
- Install that in CMS
- Add any Video or Problem block and you should see the configurations form
- Add values and save It should save successfully
- Refresh and check the values retain
- Edit any configuration in the form and save again and check that it persists


src/ol_openedx_chat/block.py
```diff
@@ -36,22 +40,70 @@ class OLChatAside(XBlockAside):
...
+        default="",
+        scope=Scope.content,
+        help=_("Prompt hint text for chat in a block"),
+    )
+    additional_solution = String(
+        display_name=_("Additional solution for problem"),
+        default="",
+        scope=Scope.content,
+        help=_("Additional solution for the problem in context of chat"),
+    )
```
asadali145, line 60: Is this field going to be the answer to the prompts?

src/ol_openedx_chat/block.py
```diff
@@ -36,22 +40,70 @@ class OLChatAside(XBlockAside):
...
+        default=False,
+        scope=Scope.content,
+        help=_("Indicates whether or not Open Learning chat is enabled for a block"),
+    )
+    chat_prompts = String(
+        display_name=_("Open Learning Chat Prompt text"),
+        default="",
+        scope=Scope.content,
+        help=_("Prompt hint text for chat in a block"),
+    )
```
asadali145, line 54: How are we going to handle multiple prompts? Comma separated maybe OR some other separator?

src/ol_openedx_chat/settings/common.py
```diff
@@ -1,11 +1,12 @@
...
     """
-    Populate settings
+    Populate common settings
     """
-
-
-DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
+    env_tokens = getattr(settings, "ENV_TOKENS", {})
+    settings.OL_CHAT_LLM_MODELS = env_tokens.get("OL_CHAT_LLM_MODELS", [])
+    settings.OL_CHAT_API_KEY = env_tokens.get("OL_CHAT_API_KEY", [])
```
asadali145, line 12: Different models will have different keys. We will have to maintain the correct order in lists otherwise we can use the dicts.

asadali145 (approved):
LGTM! I have left some questions that came to mind while reviewing.

src/ol_openedx_chat/README.rst
```diff
@@ -1,23 +1,81 @@
...
+
+   ::
+
+      Follow these steps in a terminal on your machine:
+
+      1. Navigate to open-edx-plugins directory
+
+      2. If you haven't done so already, run "./pants build"
+
+      3. Run "./pants package ::". This will create a "dist" directory inside "open-edx-plugins" directory with ".whl" & ".tar.gz" format packages for "ol_openedx_chat" and other "ol_openedx_*" plugins in "open-edx-plugins/src"
```
asadali145, line 47: This will be `pants` instead of `./pants`. Right?

src/ol_openedx_chat/README.rst
```diff
@@ -1,23 +1,81 @@
...
+
+      4. Move/copy any of the ".whl" or ".tar.gz" files for this plugin that were generated in the above step to the machine/container running Open edX (NOTE: If running devstack via Docker, you can use "docker cp" to copy these files into your LMS or CMS containers)
+
```
asadali145, line 50: Another option with the Devstack, if open-edx-plugins is in the src folder, we can go to shell and do `pip install /edx/src/open-edx-plugins/dist/ol-chat...tar.gz`

asadali145 (approved):
LGTM!
Tests are failing and I am not sure why.

asadali145 (approved):

PR #398, opened 2024-11-21
feat: add basic lms chat

asadali145:

Closes https://github.com/mitodl/hq/issues/6087

Adds a basic chat window in the LMS for the MCQ blocks.

https://github.com/user-attachments/assets/d6e650c9-6f6b-436d-b71c-dbb5a3753b03

### How can this be tested?

- Build it locally by `pants package :: .`
- Install in LMS by going to the lms shell `pip install /edx/src/open-edx-plugins/dist/ol-openedx-chat-0.1.0.tar.gz`
- Open any MCQ problem

### DIFF
--- a/src/ol_openedx_chat/static/js/ai_chat.js
+++ b/src/ol_openedx_chat/static/js/ai_chat.js
@@ -1,4 +1,4 @@
-function AiChatAsideView(runtime, element) {
+function AiChatAsideView(runtime, element, block_element, init_args) {
     $('.chat-button').on('click', function () {
         const blockKey = $(this).data("block-key")
         const chatWindowSelector = '#chat-window-' + blockKey
@@ -72,4 +72,5 @@ function AiChatAsideView(runtime, element) {
             sendMessage();
         }
     });
+    console.log(init_args)
 }
--- a/src/ol_openedx_chat/static/js/ai_chat.js
+++ b/src/ol_openedx_chat/static/js/ai_chat.js
@@ -1,13 +1,19 @@
 function AiChatAsideView(runtime, element) {
     $('.chat-button').on('click', function () {
-        $('.chat-window').fadeIn();
-        $('.chat-button').hide();
+        const blockKey = $(this).data("block-key")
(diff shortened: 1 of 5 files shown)


src/ol_openedx_chat/static/js/ai_chat.js
```diff
@@ -0,0 +1,75 @@
...
+            //     // Scroll to the bottom of the chat body
+            //     $('.chat-body').scrollTop($('.chat-body')[0].scrollHeight);
+            // }, 1000);
+
+            // Scroll to the bottom after user message
+            $('.chat-body').scrollTop($('.chat-body')[0].scrollHeight);
+
+            $.ajax({
+                type: "POST",
+                url: runtime.handlerUrl(element, 'mock_handler'),
```
ChristopherChudzicki, line 48: **Question:**  Does tthis call the problem-specific `OLChatAside` mock_handler method? So the actual URL might be like `problem/uuid-1234-xyz/`?

And is this way, the chat API knows the context—which problem is it helping with?

Am I understanding that roughly correctly?

src/ol_openedx_chat/static/js/ai_chat.js
```diff
@@ -0,0 +1,75 @@
+function AiChatAsideView(runtime, element) {
+    $('.chat-button').on('click', function () {
+        const blockKey = $(this).data("block-key")
```
ChristopherChudzicki, line 3: Minor question... I see the `data-` attribute can be used to pass serialized data from python to the JS file. 

I noticed in https://docs.openedx.org/_/downloads/xblock/en/latest/pdf/#page=60 the JS function had signature
```js
function ThumbsAside(runtime, element, block_element, init_args) { /* etc */ }
```
Do we control the 3rd argument? Is that another way to pass data?

src/ol_openedx_chat/static/js/ai_chat.js
asadali145: Yes, You are spot on.

PR #408, opened 2024-12-03
fix: move from nightly to main branch

arslanashraf7:
None, The CI was broken and this is an attempt to fix that

Tutor recently renamed Its branches, Specifically the `nightly` branch that we were using is now `main`. More details are in https://github.com/overhangio/tutor/pull/1153

### How can this be tested?
All the checks should pass


Anas12091101 (approved):
LGTM 👍

PR #422, opened 2025-01-08
feat: AI Chat drawer

asadali145:

Closes https://github.com/mitodl/hq/issues/6514

Adds chat AI button when it is enabled for a problem or a video block. Also, fixes an issue with the Aside field naming + adds a new field for the drawer title.

| Desktop Non-Staff User | Desktop Staff User |
|--------|--------|
| desktop-normal | desktop-staff | 

| Tablet Non-Staff User | Tablet Staff User |
|--------|--------|
| tablet-normal | tablet-staff | 

| Mobile Non-Staff User | Mobile Staff User |
|--------|--------|
| mobile-normal | mobile-staff | 

### How can this be tested?

- Generate package by checking out this branch and running `pants package :: .`
- Enable xBlock configurations in CMS <STUDIO_URL>/admin/xblock_config/studioconfig/
- Enable xBlock asides in LMS <LMS_URL>/admin/lms_xblock/xblockasidesconfig/
- Install the ol-openedx-chat in LMS and CMS, Reload them
- In frontend-app-learning:
  - Create env.config.jsx and add the below code:
```
import * as remoteAiChatDrawer from "./mitodl-smoot-design/dist/bundles/remoteAiChatDrawer.umd.js"

remoteAiChatDrawer.init({
  messageOrigin: "http://local.openedx.io:8000",
  transformBody: messages => ({ message: messages[messages.length - 1].content }),
})

const config = {
  ...process.env,
};

export default config;
```
- Now run the below in the shell inside the learning MFE folder:
- `npm pack @mitodl/smoot-design@3.4.0`
- `tar -xvzf mitodl-smoot-design-3.4.0.tgz`
- `mv package mitodl-smoot-design`
- Now start learning MFE by `npm run dev`
- Visit the CMS and enable the chat for a few problem and video blocks.
- Visit LMS and you should see the button for AI Chat drawer and clicking on the button should open a drawer.
- To communicate with the LEARN AI backend, you can add the API URL in LMS private.py: `LEARN_AI_API_URL = "http://ai.open.odl.local:8002/http/recommendation_agent/"`


pdpinch:
will this close https://github.com/mitodl/hq/issues/6514

asadali145:
> will this close [mitodl/hq#6514](https://github.com/mitodl/hq/issues/6514)

Yes, It will. I can add the issue in the PR description.

src/ol_openedx_chat/static/js/ai_chat.js
```diff
@@ -0,0 +1,35 @@
...
+          {
+            type: "smoot-design::chat-open",
+            payload: {
+              askTimTitle: event.data.askTimTitle,
+              apiUrl: init_args.learn_ai_api_url,
+              initialMessages: INITIAL_MESSAGES,
+              conversationStarters: starters,
+            },
+          },
+          "http://apps.local.openedx.io:2000", // Ensure correct parent origin
```
ChristopherChudzicki, line 24: Can this be an env var?

asadali145: Good catch, Let me fix this.

src/ol_openedx_chat/static/js/ai_chat.js
```diff
@@ -0,0 +1,35 @@
...
+        },
+      ];
+      $(`#chat-button-${init_args.block_usage_key}`).on("click", { starters: init_args.starters, askTimTitle: init_args.ask_tim_drawer_title }, function (event) {
+        const starters = event.data.starters.map(message => ({ content: message }));
+
+        window.parent.postMessage(
+          {
+            type: "smoot-design::chat-open",
+            payload: {
+              askTimTitle: event.data.askTimTitle,
```
ChristopherChudzicki, line 18: I haven't tried this as I'm currently refreshing edx-platform (oops). But:

**Suggestion:** Add `chatId: <blockKey / uuid>` here. Currently I believe each time you open the drawer it's a new chat (as far as the UI is concerned) so you can't see previous messages.

With the chatId, you should be able to open problem 1, then problem 2, then problem 1, and see the chat history.

src/ol_openedx_chat/static/html/student_view.html
```diff
@@ -1,3 +1,11 @@
-<div>
-      <button type="button" onclick="alert('Hello World!')">Hello World!</button>
+<div class="{{ block_type }}-block-chat-button-container">
+  <div class="ai-chat-button {{ block_type }}-block-chat-button" id="chat-button-{{ block_key }}" data-block-key="{{ block_key }}">
+    {% if block_type == "problem" %}
+      <p>Help me with this problem</p>
+    {% elif block_type == "video" %}
+      <p><b>Questions about this video?</b><br>AskTim</p>
+    {% endif %}
+    <img src="/static/images/icon.svg" alt="Chat Icon" class="chat-icon">
```
ChristopherChudzicki, line 8:
```suggestion
    <img src="/static/images/icon.svg" alt="" class="chat-icon">
```

The icon is purely decorative. Better for screenreaders to ignore it. The button text is all that screenreaders need.

ChristopherChudzicki (approved):
This looks good! Tested a few problems using `LEARN_AI_API_URL="http://ai.open.odl.local:8002/http/recommendation_agent/"` 

One non-blocking issue I did notice was the studio UI workflow seems a little odd. Maybe this is a quirk of xblock asides in studio:

Screenshot 2025-02-20 at 11 02 06 AM

Excited to see this on RC!

asadali145:
> One non-blocking issue I did notice was the studio UI workflow seems a little odd. Maybe this is a quirk of xblock asides in studio:

I noticed this as well. I can work on this in a follow-up PR while we test it out on RC. I can create an issue for this tomorrow.

src/ol_openedx_chat/block.py
```diff
@@ -79,7 +86,45 @@ def student_view_aside(self, block, context=None):
...
+            "block_usage_key": self.scope_ids.usage_id.usage_key.block_id,
+            "user_id": self.runtime.user_id,
+            "learn_ai_api_url": settings.LEARN_AI_API_URL,
+            "learning_mfe_base_url": settings.LEARNING_MICROFRONTEND_URL,
+        }
+
+        if getattr(block, "category", None) == "video":
+            try:
+                content, filename, mimetype = get_transcript_from_contentstore(
+                    block, "en", "txt", block.get_transcripts_info()
```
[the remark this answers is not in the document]
asadali145: We are not using the transcript for now. It's better to remove this as we will pass the block ID and the transcript will be fetched separately. I can remove this.

src/ol_openedx_chat/block.py
```diff
@@ -64,6 +65,12 @@ class OLChatAside(XBlockAside):
         scope=Scope.content,
         help=_("Selected LLM model to be used for a block"),
     )
+    ask_tim_drawer_title = String(
+        display_name=_("Open Learning Drawer Title"),
+        default="",
+        scope=Scope.content,
+        help=_("Drawer title displayed in the chat drawer"),
+    )
```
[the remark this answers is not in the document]
asadali145: I have followed the existing field pattern that you added. We haven't used enforced_type + the settings scope. Looking at what it means makes me think that we should set the settings scope for all other fields as well.

> Also, Have we tested this if the title is too long? Would that break the design?

This will be passed to the drawer from the smooth design. I can test it but it is not related to the current PR.

> but I think we should set about {block.display_name} from below code as a default here

We don't have a block here it self, I am not sure of a way to set the default here. That is why I used the defaults when displaying in CMS/LMS.

src/ol_openedx_chat/block.py
```diff
@@ -79,7 +86,45 @@ def student_view_aside(self, block, context=None):
...
+
+        if getattr(block, "category", None) == "video":
+            try:
+                content, filename, mimetype = get_transcript_from_contentstore(
+                    block, "en", "txt", block.get_transcripts_info()
+                )
+            except Exception:  # noqa: BLE001
+                content = ""
+
+            extra_context["video_transcript"] = content
```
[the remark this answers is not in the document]
asadali145: This can be removed as I mentioned above.

PR #429, opened 2025-01-23
Allowed urls updated to bypass asset routes

marslanabdulrauf:
https://github.com/mitodl/hq/issues/6479#issuecomment-2606061294

This PR adds [new routes](https://github.com/openedx/edx-platform/blob/open-release/sumac.master/openedx/core/djangoapps/contentserver/urls.py) added for contentserver after [this implementation](https://github.com/openedx/edx-platform/issues/34702)

### How can this be tested?
1. Add Course assets from studio
2. Copy the web url
3. Open it in a new tab/incognito tab without login
4. It should be accessible if the asset is unlocked/public

### Additional Context
If the course asset is locked, it shouldn't be accessible

### DIFF
--- a/src/openedx_companion_auth/BUILD
+++ b/src/openedx_companion_auth/BUILD
@@ -12,7 +12,7 @@ python_distribution(
     ],
     provides=setup_py(
         name="openedx-companion-auth",
-        version="1.1.0",
+        version="1.1.1",
         description="A package to add login redirection from Open edX to MIT applications",
         license="BSD-3-Clause",
         author="MIT Office of Digital Learning",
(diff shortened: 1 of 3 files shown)


pdpinch:
I'd like to make sure that this change doesn't affect the behavior of assets that are "locked" in studio. They should still require login. I presume this access control will be managed by edx-platform, not this middleware.

marslanabdulrauf:
> I'd like to make sure that this change doesn't affect the behavior of assets that are "locked" in studio. They should still require login. I presume this access control will be managed by edx-platform, not this middleware.

yes, if an asset is locked it will still require login and that behaviour is being control by edx-platform

marslanabdulrauf:
Removing this PR as the routes are dynamically updated in `ol-infrastructure`: https://github.com/mitodl/ol-infrastructure/pull/2917

PR #447, opened 2025-03-03
feat: add block usage key and transcript asset id

asadali145:

Closes https://github.com/mitodl/hq/issues/6843

Passes the block usage key and transcripts asset ID.

Here are the example block usage keys:
```
Video: block-v1:testx+test101+2024_T1+type@video+block@d6d49d1d72f942cd9d534a54f99d8a87
Dropdown: block-v1:testx+test101+2024_T1+type@problem+block@a5e043d9325e4ed9b8912d1131051321
MCQ: block-v1:testx+test101+2024_T1+type@problem+block@be69872810904ad195e8ec9241a59e5d
```
Here is an example Asset ID for the video block English transcript:
`asset-v1:testx+test101+2024_T1+type@asset+block@91204cd0-44cb-4b68-9e44-303ff8ffc2e3-en.srt`

### How can this be tested?

- Generate package by checking out this branch and running `pants package :: .`
- Enable xBlock configurations in CMS <STUDIO_URL>/admin/xblock_config/studioconfig/
- Enable xBlock asides in LMS <LMS_URL>/admin/lms_xblock/xblockasidesconfig/
- Install the ol-openedx-chat in LMS and CMS, Reload them
  - make sure to install the latest version 0.2.2
- In frontend-app-learning:
  - Create env.config.jsx and add the below code:
```
import * as remoteAiChatDrawer from "./mitodl-smoot-design/dist/bundles/remoteAiChatDrawer.umd.js"

remoteAiChatDrawer.init({
  messageOrigin: "http://local.openedx.io:8000",
  transformBody: messages => ({ message: messages[messages.length - 1].content }),
})

const config = {
  ...process.env,
};

export default config;
```
- Now run the below in the shell inside the learning MFE folder:
- `npm pack @mitodl/smoot-design@3.4.0`
- `tar -xvzf mitodl-smoot-design-3.4.0.tgz`
- `mv package mitodl-smoot-design`
- Now start learning MFE by `npm run dev`
- Now enable the `ol_openedx_chat.ol_openedx_chat_enabled` waffle flag for a course at <LMS_URL>/admin/waffle_utils/waffleflagcourseoverridemodel/
- Verify that the chat options are only enabled for this course and are disabled for all others.
- Visit the CMS and enable the chat for a few problem and video blocks.
- For the video block:
  - Click on Edit button at the right top of the block in CMS
  Screenshot 2025-03-03 at 10 02 27 PM
  - Click on `import youtube transcript`.
  Screenshot 2025-03-03 at 10 02 47 PM
  - When we fetch transcripts, it returns an empty list. We will have to manually upload transcripts.
  - Click on the `Download Transcript for Editing`
  Screenshot 2025-03-03 at 10 03 12 PM
  - Now upload the downloaded file
  Screenshot 2025-03-03 at 10 03 43 PM
  - Save the settings, enable the AI Assistant, and Publish the unit.
- Visit LMS and you should see the button for AI Chat drawer and clicking on the button should open a drawer.
- Block usage key and Asset ID are passed to the drawer. You can check by adding a console log in ai_chat.js at line 16
  - `console.log(event.data.requestOpts)`
- To communicate with the LEARN AI backend, you can add the API URL in LMS private.py: `LEARN_AI_API_URL = "http://ai.open.odl.local:8002/http/recommendation_agent/"`

### DIFF
--- a/src/ol_openedx_chat/block.py
+++ b/src/ol_openedx_chat/block.py
@@ -1,3 +1,5 @@
+import logging
+
 import pkg_resources
 from django.conf import settings
 from django.template import Context, Template
@@ -11,6 +13,8 @@ from xblock.fields import Boolean, Scope
 from xmodule.x_module import AUTHOR_VIEW, STUDENT_VIEW
 
 from .compat import get_ol_openedx_chat_enabled_flag
+
+log = logging.getLogger(__name__)
 
 
 def get_resource_bytes(path):
@@ -54,6 +58,7 @@ class OLChatAside(XBlockAside):
         """
         Renders the aside contents for the student view
         """  # noqa: D401
+        from xmodule.video_block.transcripts_utils import Transcript
 
         # This is a workaround for those blocks which do not have has_author_view=True
         # because when a block does not define has_author_view=True in it, the only view
@@ -78,10 +83,29 @@ class OLChatAside(XBlockAside):
         )
         fragment.add_css(get_resource_bytes("static/css/ai_chat.css"))
         fragment.add_javascript(get_resource_bytes("static/js/ai_chat.js"))
+
+        transcript_asset_id = None
+        if getattr(block, "category", None) == "video":
+            try:
+                transcripts_info = block.get_transcripts_info()
+                if transcripts_info.get("transcripts") and transcripts_info[
+                    "transcripts"
+                ].get("en"):
+                    transcript_asset_id = Transcript.asset_location(
+                        block.location, transcripts_info["transcripts"]["en"]
+                    )
+
+            except Exception:
+                log.exception(
+                    "Error while fetching transcripts for block %s",
+                    block.location,
+                )
+
         extra_context = {
             "ask_tim_drawer_title": f"about {block.display_name}",
             "block_id": self.scope_ids.usage_id.usage_key.block_id,
             "block_usage_key": self.scope_ids.usage_id.usage_key,
+            "transcript_asset_id": transcript_asset_id,
             "user_id": self.runtime.user_id,
             "learn_ai_api_url": settings.LEARN_AI_API_URL,
             "learning_mfe_base_url": settings.LEARNING_MICROFRONTEND_URL,
--- a/src/ol_openedx_chat/block.py
+++ b/src/ol_openedx_chat/block.py
@@ -89,8 +89,7 @@ class OLChatAside(XBlockAside):
         fragment.add_css(get_resource_bytes("static/css/ai_chat.css"))
         fragment.add_javascript(get_resource_bytes("static/js/ai_chat.js"))
 
-        request_opts = {
-            "block_id": block_id,
+        request_body = {
             "block_usage_key": block_usage_key,
         }
 
@@ -100,13 +99,13 @@ class OLChatAside(XBlockAside):
                 if transcripts_info.get("transcripts") and transcripts_info[
                     "transcripts"
                 ].get("en"):
-                    request_opts["transcript_asset_id"] = Transcript.asset_location(
+                    request_body["transcript_asset_id"] = Transcript.asset_location(
                         block.location,
                         transcripts_info["transcripts"][ENGLISH_LANGUAGE_TRANSCRIPT],
                     )
 
-            except Exception:
-                log.exception(
+            except Exception:  # noqa: BLE001
+                log.info(
                     "Error while fetching transcripts for block %s",
                     block.location,
                 )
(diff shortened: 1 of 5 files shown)


src/ol_openedx_chat/block.py
```diff
@@ -66,23 +72,51 @@ def student_view_aside(self, block, context=None):
...
+                transcripts_info = block.get_transcripts_info()
+                if transcripts_info.get("transcripts") and transcripts_info[
+                    "transcripts"
+                ].get("en"):
+                    request_opts["transcript_asset_id"] = Transcript.asset_location(
+                        block.location,
+                        transcripts_info["transcripts"][ENGLISH_LANGUAGE_TRANSCRIPT],
+                    )
+
+            except Exception:
```
pdpinch, line 108: It's possible that the video might not have any transcripts, or no English language transcript. For example, it might be a demonstration video without audio. 

Warning on missing transcripts should be OK, but I don't think we want to raise an exception.

ChristopherChudzicki (changes requested):
This worked well for me. I have one small rename request if you agree with it.

Additionally: we should verify with whoever is building the Video chatbot (@arslanashraf7 , I think?) that `{ block_id, block_usage_key, transcript_asset_id }` are good names for the API to use. Seems OK to me.

src/ol_openedx_chat/block.py
```diff
@@ -66,23 +72,51 @@ def student_view_aside(self, block, context=None):
...
+                    block.location,
+                )
+
         extra_context = {
             "ask_tim_drawer_title": f"about {block.display_name}",
-            "block_usage_key": self.scope_ids.usage_id.usage_key.block_id,
             "user_id": self.runtime.user_id,
             "learn_ai_api_url": settings.LEARN_AI_API_URL,
             "learning_mfe_base_url": settings.LEARNING_MICROFRONTEND_URL,
+            "request_opts": request_opts,
```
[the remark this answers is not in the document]
asadali145: Sounds good! I can rename it.

src/ol_openedx_chat/block.py
```diff
@@ -66,23 +72,51 @@ def student_view_aside(self, block, context=None):
         if not self.ol_chat_enabled:
             return fragment
 
+        block_id = self.scope_ids.usage_id.usage_key.block_id
+        block_usage_key = self.scope_ids.usage_id.usage_key
```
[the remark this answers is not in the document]
asadali145: Yes, it's all the same. The class is named [BlockUsageLocator](https://github.com/openedx/opaque-keys/blob/master/opaque_keys/edx/locator.py#L621), so we generally use variable names like usage_key.

asadali145:
> Additionally: we should verify with whoever is building the Video chatbot (@arslanashraf7, I think?) that { block_id, block_usage_key, transcript_asset_id } are good names for the API to use. Seems OK to me.

I have removed the `block_id` from the `requestBody`, Now, we will have `block_usage_key` and `transcript_asset_id`.

asadali145:
@arslanashraf7 thanks, i can create the issues you mentioned. The PR should be good now as per your suggestions.

src/ol_openedx_chat/block.py
```diff
@@ -66,23 +71,51 @@ def student_view_aside(self, block, context=None):
...
+                    "block_type": block_type,
                 },
             )
         )
         fragment.add_css(get_resource_bytes("static/css/ai_chat.css"))
         fragment.add_javascript(get_resource_bytes("static/js/ai_chat.js"))
+
+        request_body = {
+            "edx_module_id": block_usage_key,
+        }
```
ChristopherChudzicki, line 93: I agree with @arslanashraf7 's suggestion about using naming to align with learn so I think `edx_module_id` is good.

But we do want to send the transcript module_id also.

Maybe send `video_edx_module_id` and `transcript_edx_module_id`

?

asadali145: > Maybe send video_edx_module_id and transcript_edx_module_id

edx_module_id is generic for videos and problems. I think `transcript_asset_id ` is good for now.

sovsey:
> If the video doesn't have transcripts, Should the chat widget be visible? I ask this because if there is no transcript the video GPT would not be able to answer questions.

If there aren't transcripts, can we hide the VideoGPT button? Thoughts @pdpinch ?

ChristopherChudzicki (approved):

src/ol_openedx_chat/block.py
ChristopherChudzicki: sorry—this was a mistaken comment on my part. I forgot/missed during my re-review that we add the transcript id below (of course, since it depends on block type).

ChristopherChudzicki (approved):

PR #454, opened 2025-03-11
feat: pass block type to chat drawer

asadali145:

None

Passes block type to the AI Chat drawer.

### How can this be tested?

- Setup by following the [Readme](https://github.com/mitodl/open-edx-plugins/tree/main/src/ol_openedx_chat).
- You can add console log in ai_chat.js to see what is available in event.data.blockType.


ChristopherChudzicki (approved):
👍

PR #459, opened 2025-03-21
tests: add tests for ol openedx chat

asadali145:

Closes https://github.com/mitodl/hq/issues/6935

Adds tests for the ol_openedx_chat plugin.

### How can this be tested?

- Checks should pass and ol_openedx_chat should work as expected.

### DIFF
--- a/run_devstack_integration_tests.sh
+++ b/run_devstack_integration_tests.sh
@@ -63,6 +63,11 @@ for subdir in "src"/*; do
                 echo "No tests found, skipping pytest."
             fi
 
+            # Run the pytest command with CMS settings (for ol_openedx_chat)
+            if [[ "$subdir" == *"ol_openedx_chat"* ]]; then
+                pytest . --cov . --ds=cms.envs.test
+            fi
+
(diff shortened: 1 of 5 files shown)


asadali145:
@arslanashraf7 This is rebased and should be good for review.

asadali145:
> In my opinion we have both LMS and CMS setup through tutor while running tests so they should not be skipped.

We are already doing that. As you see I have updated the scripts to run the ol_openedx_chat with both LMS and CMS.

if you scroll down in the GitHub action logs, you will see the tests running with CMS settings.

Running with LMS:
Screenshot 2025-03-28 at 9 13 56 PM

Running with CMS:
Screenshot 2025-03-28 at 9 14 27 PM


We need to run a few tests in LMS, a few in CMS, and a few in both LMS and CMS. I have used the decorators `skip_unless_cms` and `skip_unless_lms` to achieve that.

asadali145:
So, a few tests will be skipped when running with LMS and a few will skip when running with CMS.

PR #495, opened 2025-05-07
fix(git-auto-export): Start export only if GIT_URL exists in the course settings

marslanabdulrauf:
https://github.com/mitodl/hq/issues/6803

This PR:
1. Start course export on publish only when giturl exists

### How can this be tested?
1. Checkout to this branch
2. Create installable package by following the instructions in [Readme file](https://github.com/mitodl/open-edx-plugins/tree/main/src/ol_openedx_git_auto_export)
3. Install the package in edX.
4. Create a course in studio and make sure giturl value is set to empty or null in course advance settings
5. Publish the course
6. Keep an eye on logs -- you should see `Course {course_key} does not have a giturl, skipping export.` instead of error saying `'NoneType' object has no attribute 'endswith'`

### DIFF
--- a/src/ol_openedx_git_auto_export/BUILD
+++ b/src/ol_openedx_git_auto_export/BUILD
@@ -12,7 +12,7 @@ python_distribution(
     dependencies=[":git_auto_export"],
     provides=setup_py(
         name="ol-openedx-git-auto-export",
-        version="0.3.1",
+        version="0.3.2",
         license="BSD-3-Clause",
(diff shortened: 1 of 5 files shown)


marslanabdulrauf:
I did a basic search in [ol-infrastructur](https://github.com/search?q=repo:mitodl/ol-infrastructure%20GIT_REPO_EXPORT_DIR&type=code) and they are using `/openedx/data/export_course_repos` which is safe.

PR #496, opened 2025-05-07
chore(edx-sysadmin): Improve logging when incoming branch conflicts with setting branch

marslanabdulrauf:
https://github.com/mitodl/ol-infrastructure/issues/2891#issuecomment-2855121735

This PR:
1. Logs exception only when necessary

### How can this be tested?
1. Checkout to this branch
2. Create installable package by following the instructions in [Readme file](https://github.com/mitodl/open-edx-plugins/tree/main/src/edx_sysadmin)
3. Install the package in edX
4. Make sure you have setup the hook for course auto reload
5. Change the default branch setting for this plugin to something different. For example: `dummy`
6. Publish any course from studio without any change which will invoke the hook for edx_sysadmin > gitreload OR directly push changes to github repo whose hook is configured
7. Verify that there are debug logs like `Couldn't reload course for the branch` instead of exception

### DIFF
--- a/src/edx_sysadmin/BUILD
+++ b/src/edx_sysadmin/BUILD
@@ -21,7 +21,7 @@ python_distribution(
     ],
     provides=setup_py(
         name="edx-sysadmin",
-        version="0.3.0",
+        version="0.3.1",
         description="An Open edX plugin to enable SysAdmin panel",
         license="BSD-3-Clause",
         author="MIT Office of Digital Learning",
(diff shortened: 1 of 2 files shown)


pdpinch:
Since this is a monorepo, please update the PR title and the commit message with some more detail about which plugin's logging was improved.

marslanabdulrauf:
Message example: `Couldn't reload course from the branch (live), expected branch was (main)`

PR #513, opened 2025-06-02
chore: update version git auto export

asadali145:

None

Updates version of git auto export to include the updated dependency range for opaque keys after merging https://github.com/mitodl/open-edx-plugins/pull/491


asadali145:
Merging because we don't have tests for this.

PR #546, opened 2025-07-07
refactor: update rapid response to use studio runtime

arslanashraf7:
https://github.com/mitodl/hq/issues/7737

- Updates the rapid response xBlock to use the studio runtime. This xBlock aside was using an alternate API view-based mechanism in the past to enable/disable the rapid response based on the toggle.
- In addition to the old code, when the rapid response was automatically published as well upon toggle, this time we will hit the cms update API, and the toggle will now behave the same as the built-in xblocks do. e.g., Upon clicking the toggle the `Publish` button would be enabled.

https://github.com/user-attachments/assets/d786876c-66b1-4003-8e18-2466b861180c

### How can this be tested?
- Setup the Rapid response from this branch
- Enable/Disable the rapid response, should `Enable` the publish button instead of publishing the rapid response by default under the hood
- Refresh the page and check that the toggle state is retained
- Now publish the XBlock unit. (If the rapid response was enabled, then you should see the rapid response graph on LMS, and if it was disabled, you should see nothing related to rapid response on the LMS side)

### DIFF
--- a/src/rapid_response_xblock/rapid_response_xblock/static/js/src_js/rapid_studio.js
+++ b/src/rapid_response_xblock/rapid_response_xblock/static/js/src_js/rapid_studio.js
@@ -5,7 +5,7 @@
     // we need studio runtime to get handler capable of saving xblock data in Studio
     var studioRuntime = new window.StudioRuntime.v1();
 
-    var toggleEnabledUrl = studioRuntime.handlerUrl(element, 'toggle_block_enabled');
+    var toggleUrl = studioRuntime.handlerUrl(element, 'toggle_block_enabled');
     var $element = $(element);
 
     var rapidTopLevelSel = '.rapid-response-block';
@@ -28,7 +28,7 @@
               message: gettext('Toggling rapid response')
           });
 
-          $.post(toggleEnabledUrl).then(
+          $.post(toggleUrl).then(
             function(state) {
               render(state);
             }


src/rapid_response_xblock/static/js/src_js/rapid_studio.js
```diff
@@ -2,30 +2,44 @@
...
-    var toggleEnabledUrl = runtime.handlerUrl(element, '');
-    // Redirect this call to our own API instead of xBlock handler because the handler callbacks are not supported
-    // for the studio preview. The calls can have two formats 1) "preview/xblock" when called from studio preview
-    // 2) "/xblock" when called from Studio plugin(Tab) settings.
-    toggleEnabledUrl = toggleEnabledUrl.replace("preview/xblock", "toggle-rapid-response")
-    toggleEnabledUrl = toggleEnabledUrl.replace("/xblock", "/toggle-rapid-response")
+    // we need studio runtime to get handler capable of saving xblock data in Studio
+    var studioRuntime = new window.StudioRuntime.v1();
+
+    var toggleEnabledUrl = studioRuntime.handlerUrl(element, 'toggle_block_enabled');
```
asadali145, line 8: nit: 
```suggestion
    var toggleUrl = studioRuntime.handlerUrl(element, 'toggle_block_enabled');
```

PR #554, opened 2025-07-14
Add social auth debugging actions

rhysyngsun:

N/A

This adds a social auth pipeline action that is a copy of the version from the core social auth library but adds some logging to try to help us debug the state conflicts.

### How can this be tested?

Recommend you install this plugin by following the instructions in `docs/README.rst` to mount this repo into the LMS/CMS. I then ran:

```bash
tutor config save \
  --remove OPENEDX_EXTRA_PIP_REQUIREMENTS="ol-social-auth" \
  --append OPENEDX_EXTRA_PIP_REQUIREMENTS="-e /openedx/open-edx-plugins/src/ol_social_auth"
```

And rebuilt openedx images and restarted the lms and cms per the README.

Then in order to test this, you have to artificially create a situation where this happens. The easiest way to do this is to take a user that you can currently successfully login with, go to their UserSocialAuth record and change the user to be someone else.


src/ol_social_auth/tests/test_actions.py
```diff
@@ -0,0 +1,53 @@
...
+from social_core.exceptions import AuthAlreadyAssociated
+from social_django.models import UserSocialAuth
+
+from ol_social_auth.actions import debug_social_user
+from ol_social_auth.backends import OLOAuth2
+
+pytestmark = pytest.mark.django_db
+
+
+@skip_unless_cms
```
asadali145, line 13: This will skip the tests when running with LMS settings. If you want to run these tests in the CMS environment you will have to change the below code block in tests script:

```
    # Run the pytest command with CMS settings (for ol_openedx_chat)
    if [[ "$plugin_dir" == *"ol_openedx_chat"* || "$plugin_dir" == *"ol_openedx_course_sync"* ]]; then
        pytest . --cov . --ds=cms.envs.test

        PYTEST_SUCCESS=$?
        if [[ $PYTEST_SUCCESS -ne 0 ]]
        then
            echo "pytest exited with a non-zero status"
            exit $PYTEST_SUCCESS
        fi
    fi
``` 

Previously, we only had tests for LMS and now recently we added tests that we need to run with CMS settings so introduced this thing in the script.

Otherwise, we will have to add the decorator on all old tests to skips them for CMS.

rhysyngsun:
Closing this as we solved it without needing this.

PR #558, opened 2025-07-15
fix: chat xblock indexing error

arslanashraf7:
https://github.com/mitodl/hq/issues/7813

We were seeing indexing error on the chat xBlock, it would only pop up when we click the `Re-index` button on the CMS/Authoring course outline page. They did not block the actual testing but the indexing itself.

- [ ] Desktop screenshots

### How can this be tested?

- On your existing setup locally, Run the chat xblock v0.1.0
- Add some chat xblocks in the unit
- Publish them in CMS
- click `Re-index` on the course outline page (IT will break)
- Now shift to this branch, v0.1.1 and the same steps should not break anything.

### DIFF
--- a/src/ol_openedx_chat_xblock/block.py
+++ b/src/ol_openedx_chat_xblock/block.py
@@ -4,7 +4,6 @@ import pkg_resources
 import requests
 from django.conf import settings
 from django.template import Context, Template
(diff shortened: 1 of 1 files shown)


asadali145 (approved):
LGTM!

PR #566, opened 2025-07-17
increase chat height

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/7849

Increases height of canvas chat ui.

**In Canvas:** (Manually edited the existing CSS—i.e., not actually this version—but the canvas view is what matters.)
Screenshot 2025-07-17 at 5 30 59 PM

**In MFE LMS**:
Screenshot 2025-07-18 at 11 16 53 AM

**In Legacy Studio:**
Screenshot 2025-07-18 at 11 06 09 AM

### How can this be tested?
1. Enable the xblock locally; view it in LMS and CMS.

@asadali145 I had trouble enabling this locally. So I actually haven't tested it locally.  I added `["ol_openedx_chat_xblock"]` to the advanced module list.

Do I need a waffle flag?

**Edit:** Resolved; tested it locally now.

### DIFF
--- a/src/ol_openedx_chat_xblock/ol_openedx_chat_xblock/static/css/ai_chat_xblock.css
+++ b/src/ol_openedx_chat_xblock/ol_openedx_chat_xblock/static/css/ai_chat_xblock.css
@@ -1,4 +1,8 @@
 .ai-chat-container {
+  /*
+  * MFE Iframe doesn't have a height set, so ensure a min height.
+  */
+  min-height: 400px;
   height: calc(100vh - 100px);
   position: relative;
 }
--- a/src/ol_openedx_chat_xblock/ol_openedx_chat_xblock/static/css/ai_chat_xblock.css
+++ b/src/ol_openedx_chat_xblock/ol_openedx_chat_xblock/static/css/ai_chat_xblock.css
@@ -1,8 +1,14 @@
 .ai-chat-container {
   /*
-  * MFE Iframe doesn't have a height set, so ensure a min height.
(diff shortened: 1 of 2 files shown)


asadali145:
Hi, have you restarted LMS and CMS after installing the plugin?

We do not have a waffle flag for the chat xBlock. But we do have a couple of settings that we should set in lms/envs/private.py and cms/envs/private.py. I'll send those over Keybase.

src/ol_openedx_chat_xblock/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "ol-openedx-chat-xblock"
-version = "0.1.2"
+version = "0.1.3"
```
asadali145, line 3: We need to update it to 0.1.4 as we already have 0.1.3

src/ol_openedx_chat_xblock/ol_openedx_chat_xblock/static/css/ai_chat_xblock.css
```diff
@@ -1,4 +1,21 @@
 .ai-chat-container {
-    min-height: 400px;
-    position: relative;
+  /*
+  * MFE Iframe doesn't have a height set, so ensure a min height.
+  */
+  min-height: 400px;
+  height: calc(100vh - 100px);
```
asadali145, line 6: I am not sure, but this is breaking Learning MFE design. Have you tested this change?

asadali145 (changes requested):

asadali145 (approved):
LGTM!

PR #572, opened 2025-07-28
feat: Auto generate the chat xBlock course ID from Canvas LTI call

arslanashraf7:
https://github.com/mitodl/hq/issues/7923

- Makes the course ID optional between xBlock settings's Course_ID and auto generated Canvas LTI call course ID.
- Adds auto generation of Canvas course ID for the xBlock based on the LTI Launch call from the Canvas LMS.
- Adds to logic to generate the new course ID
- Use this newly generated course ID for all the MIT Learn AI Canvas chat API
- Now, the chat xblock will use the course_id from xBlock settings, if not available it will try to generate the id from Canvas LTI Call.

- [ ] Desktop screenshots

### How can this be tested?
There are two ways to test this change:

1. Test with Canvas
    - Enable LTI and related configurations on the open edX side https://docs.openedx.org/en/latest/site_ops/install_configure_run_guide/configuration/lti/tpa_lti.html
    - Add the created configuration in the form of an app in https://canvas.mit.edu/courses/14566/settings/configurations
    - Make your open edX local instance public (You can use zrok, etc)
    - Create a chat xBlock, copy its URL, and convert it into an LTI URL via https://kalebabebe.github.io/mitx-canvas/vertical.html
    - Add a module in https://canvas.mit.edu/courses/14566 if you have access, and add your block there
    - Load the module in Canvas, and it should load the xBlock there. (For local instance, Canvas gives some warnings about rendering iFrame, but as long as you are able to debug the Canvas Course ID being generated, it is fine.)
    - The chat xBlock should follow https://github.com/mitodl/hq/issues/7923#issuecomment-3127113099 and generate a Course ID that learn can understand

2. Test without Canvas
    - Create a chat xBlock
    - Add the course_id to the xBlock settings as we usually do
    - Try the chat and it should work, It should not try to generate the course id from the LTI call
    - Now, Remove the course_id from xblock settings, and try chat, it should try to use the course_id generated via Canvas LTI call
    - If none is available, it should throw a proper error
    - Either mimic the canvas LTI API call or add `lti_params={"custom_course_id":"123", "context_label":"xyz"}`
    - Run the xBlock and check the generated URL 
    - You can actually test this locally if you add values in `lti_params` that actually represent a course in MIT Learn.
    -

### DIFF
--- a/src/ol_openedx_chat_xblock/ol_openedx_chat_xblock/block.py
+++ b/src/ol_openedx_chat_xblock/ol_openedx_chat_xblock/block.py
@@ -174,17 +174,17 @@ class OLChatXBlock(XBlock, StudioEditableXBlockMixin):
         # If neither is available, it will return an error response.
         course_id_for_chat = self.course_id
         if not course_id_for_chat:
-            log.error(
+            log.info(
                 "Course ID is not available in the XBlock. "
                 "Falling back to auto-generated course ID from Canvas LTI."
             )
-        if not self.learn_readable_course_id:
-            log.error("Course ID is not available from Canvas LTI.")
-            return Response(
-                "Course ID is required.",
-                status=api_status.HTTP_400_BAD_REQUEST,
-            )
-        course_id_for_chat = self.learn_readable_course_id
+            if not self.learn_readable_course_id:
+                log.error("Course ID is not available from Canvas LTI.")
+                return Response(
+                    "Course ID is required.",
+                    status=api_status.HTTP_400_BAD_REQUEST,
+                )
+            course_id_for_chat = self.learn_readable_course_id
 
         message = request_data.get("message", "").strip()
         if not message:


src/ol_openedx_chat_xblock/ol_openedx_chat_xblock/block.py
```diff
@@ -56,15 +87,21 @@ class OLChatXBlock(XBlock, StudioEditableXBlockMixin):
         scope=Scope.settings,
         help="This name appears in the horizontal navigation at the top of the page.",
     )
-    course_id = String(
+    learn_readable_course_id = String(
         default="",
-        scope=Scope.settings,
-        help="Course ID of the relevant course in Canvas",
+        scope=Scope.user_state,
```
asadali145, line 92: `learn_readable_course_id` is not related to the user. If we use `user_state`, it means we want to store user-specific data. The best is to use the settings scope. More on user_state [here](https://github.com/openedx/XBlock/blob/master/xblock/fields.py#L167).

asadali145: Thanks for letting me know that the settings scope isn't letting us edit field data.

asadali145 (approved):
LGTM! Tested locally without setting up Canvas.

pdpinch:
Thanks for bringing this up. I don't think we're quite ready to make this changeover yet. 

> We add course_id=$Canvas.course.id in the course custom fields in Canvas

Do you where/how to do this?

sovsey:
A couple questions, but generally I think we're good to merge. 

1. Will existing XBlocks work that have the ID specified? 

>If not, it will try to generate the course ID from the Canvas LTI launch call and use it.

2. Where/how would we be able to see which ID it is generating from the Canvas LTI launch?

blarghmatey:
I triggered that this morning and it is ready for testing on QA now.

PR #600, opened 2025-08-27
docs: add note for MathJax filter removal

arslanashraf7:
None, Just adds a note for future.

- Add a note that we should remove this filter when edx us also using v3 of MathJax

- [ ] Desktop screenshots


asadali145 (approved):
LGTM!

PR #603, opened 2025-09-03
chore: bump AIChat bundle to fix problem selection

arslanashraf7:
None

- Bumps the AI Chat bundle to include the fix for the problem selection issue. The issue was that the AI chat would let users send messages without selecting the problem, that would end up in error. This fixes it.
- Converts some more .error logs to .info
- Makes the chat bundle accept any updated patch versions so that we do not have to bump them every time 

https://github.com/user-attachments/assets/ca40eb03-cda8-460c-863f-21c1038de922

- [ ] Desktop screenshots

### How can this be tested?
- Setup a course with tutor XBlocks in it with the appropriate course ID set
- Reproduce the issue on the pervious chat block version
- Now, Install the chat xblock from this PR and see that the problem selection problem is fixed and the message can't be sent without selecting the problem


asadali145 (approved):
LGTM!

sovsey:
@arslanashraf7 @jonkafton how would we know when a smoot design patch release is on RC? So that we can test before Prod?

asadali145 (approved):
LGTM!

PR #615, opened 2025-09-18
feat: create github repo on course creation

marslanabdulrauf:
https://github.com/mitodl/hq/issues/6962

This PR:
1. adds receivers to create Github repo(if enabled) on course creation signal
2. creates Github repo(if enabled) on course re-run creation

### Screenshot

image

### How can this be tested?

1. Install and configure [ol_openedx_git_auto_export](https://github.com/mitodl/open-edx-plugins/tree/main/src/ol_openedx_git_auto_export)
2. Run Migration in CMS: `tutor dev exec cms ./manange.py cms migrate`

#### Testing management command
1. You must have a course with giturl set in its advance settings -- It shouldn't have gitrepo entry in [admin](http://studio.local.openedx.io:8001/admin/ol_openedx_git_auto_export/coursegitrepo/)
2. Run Management command `tutor dev exec cms ./manage.py cms migrate_giturl`
4. Verify that the giturl is migrate to [CourseGitRepo](http://studio.local.openedx.io:8001/admin/ol_openedx_git_auto_export/coursegitrepo/) model
5. Make sure that giturl is migrated for re-run courses as well

#### Auto repo creation testing
1. Install and configure [ol_openedx_git_auto_export](https://github.com/mitodl/open-edx-plugins/tree/main/src/ol_openedx_git_auto_export)
2. Make sure to:
    - Enable `ENABLE_AUTO_GITHUB_REPO_CREATION`
    - Set `GITHUB_ORG_API_URL`
    - Set `GITHUB_ACCESS_TOKEN` -- This token should have repo read/write and pull/push access.
3. Create new course and verify from your organization that a new repository is created.
4. Check [admin](http://studio.local.openedx.io:8001/admin/ol_openedx_git_auto_export/coursegitrepo/) and see if the entry is added for this newly created course
6. Create re-run of any course and repeat steps 3 and 4 to verify
7. Make any change and publish to make sure you are getting latest course changes/commits

### DIFF
--- a/src/ol_openedx_git_auto_export/pyproject.toml
+++ b/src/ol_openedx_git_auto_export/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "ol-openedx-git-auto-export"
-version = "1.0.0"
+version = "0.7.0"
 description = "A plugin that auto saves the course OLX to git when an author publishes it"
 authors = [
   {name = "MIT Office of Digital Learning"}


src/ol_openedx_git_auto_export/ol_openedx_git_auto_export/signals.py
```diff
@@ -19,32 +31,99 @@ def listen_for_course_publish(
...
-                "GIT_REPO_EXPORT_DIR is not available in settings, please create it first"  # noqa: E501
-            )
-            os.makedirs(git_repo_export_dir, 0o755)  # noqa: PTH103
-
+        get_or_create_git_export_repo_dir()
         log.info(
             "Course published with auto-export enabled. Starting export... (course id: %s)",  # noqa: E501
             course_key,
         )
+        # HACK: To create auto git repo for Re-runs as it does not emit COURSE_CREATED signal  # noqa: E501 FIX004
```
pdpinch, line 44: Is it a bug that reruns don't emit `COURSE_CREATED` signal?

Is there a different signal emitted when a rerun is created? @asadali145

pdpinch:
When we create a repo, I want to make sure that it is set to private in Github. Course exports are considered sensitive since they can contain exam solutions. 

@cachob anything else we learned recently about creating repos?

src/ol_openedx_git_auto_export/ol_openedx_git_auto_export/signals.py
asadali145: > Is there a different signal emitted when a rerun is created? @asadali145

I am not sure about a specific signal, but we listen for the post_save signal on changes in `CourseRerunState` model. It actually tracks the status of reruns, like in_progress, failed, and completed. When a rerun state is completed, then we update our course sync mappings.

src/ol_openedx_git_auto_export/ol_openedx_git_auto_export/signals.py
```diff
@@ -19,32 +31,99 @@ def listen_for_course_publish(
...
+            listen_for_course_created(course=CourseData(course_key=course_key))
 
         user = get_publisher_username(course_module)
         # If the Git auto-export is enabled, push the course changes to Git
         async_export_to_git.delay(str(course_key), user)
+
+
+def listen_for_course_created(**kwargs):
+    course_id = kwargs.get("course").course_key
+    course_id_slugified = slugify(str(course_id))
```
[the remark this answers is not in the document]
marslanabdulrauf: I tested it with all special characters that I could type from my laptop and other than `.`, `-` and `_` it converts all of them into `-`
Let me add a formatter to convert `course_id` into Github supported format, it will increase readability

src/ol_openedx_git_auto_export/ol_openedx_git_auto_export/signals.py
marslanabdulrauf: thanks @asadali145 I will update it to catch course re-run creation through `post_save` signal 👍

src/ol_openedx_git_auto_export/ol_openedx_git_auto_export/signals.py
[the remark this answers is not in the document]
pdpinch: Can we also drop the `course-v1-` from the start of the repo name? It will be the same for all repos. (unless maybe someday there is a course-v2).

marslanabdulrauf:
Thanks for clarifying @pdpinch 

I have also updated migration command to:
1. Migrate `giturl` from advance settings to our custom managed table `CourseGithubRepository`
2. Create new repo for courses with duplicate `giturl`
3. Create new repo if the course does not have any `giturl`

src/ol_openedx_git_auto_export/ol_openedx_git_auto_export/management/commands/migrate_giturl.py
```diff
@@ -0,0 +1,90 @@
...
+                    defaults={"git_url": giturl},
+                )
+            elif giturl and giturl in seen_giturls:
+                self.stdout.write(
+                    self.style.WARNING(
+                        f"Course {course.id} has a duplicate giturl: {giturl}\n"
+                        f"Creating a new GitHub repository for {course.id}"
+                    )
+                )
+                ssh_url = async_create_github_repo(course.id, export_course=True)
```
asadali145, line 76:
```suggestion
                ssh_url = async_create_github_repo(str(course.id), export_course=True)
```

src/ol_openedx_git_auto_export/ol_openedx_git_auto_export/tasks.py
```diff
@@ -20,14 +38,105 @@ def async_export_to_git(course_key_string, user=None):
             "Starting async course content export to git (course id: %s)",
             course_module.id,
         )
-        export_to_git(course_module.id, course_module.giturl, user=user)
+        course_repo = CourseGithubRepository.objects.get(course_id=course_key_string)
```
asadali145, line 41: We are overlooking the flag here and exporting in every case. Also, we should add appropriate logs.
```suggestion
        course_repo = CourseGithubRepository.objects.get(course_id=course_key_string, is_export_enabled=True)
```

src/ol_openedx_git_auto_export/ol_openedx_git_auto_export/tasks.py
```diff
@@ -20,14 +38,105 @@ def async_export_to_git(course_key_string, user=None):
...
+    if not settings.GITHUB_ORG_API_URL or not gh_access_token:
+        error_msg = "GITHUB_ORG_API_URL or GITHUB_ACCESS_TOKEN is not set in settings. Skipping GitHub repo creation."  # noqa: E501
+        raise ImproperlyConfigured(error_msg)
+
+    course_module = modulestore().get_course(course_key)
+    url = f"{settings.GITHUB_ORG_API_URL}/repos"
+    headers = {
+        "Accept": "application/vnd.github+json",
+        "Authorization": f"Bearer {gh_access_token}",
+        "X-GitHub-Api-Version": "2022-11-28",
```
asadali145, line 100: Can we move constants to constants.py?

src/ol_openedx_git_auto_export/ol_openedx_git_auto_export/tasks.py
```diff
@@ -20,14 +38,105 @@ def async_export_to_git(course_key_string, user=None):
...
+        LOGGER.info(
+            "GitHub repository created for course %s: %s",
+            course_key,
+            ssh_url,
+        )
+    else:
+        LOGGER.error(
+            "Failed to retrieve SSH URL for GitHub repository for course %s",
+            course_key,
+        )
```
asadali145, line 137: This will still go to sentry and I think we should add more information here to why we could not get the URL.

src/ol_openedx_git_auto_export/ol_openedx_git_auto_export/utils.py
```diff
@@ -19,3 +31,71 @@ def get_publisher_username(course_module):
...
+        # for development/docker/vagrant if GIT_REPO_EXPORT_DIR folder does not exist then create it  # noqa: E501
+        log.error(
+            "GIT_REPO_EXPORT_DIR is not available in settings, please create it first"
+        )
+        os.makedirs(git_repo_export_dir, 0o755)  # noqa: PTH103
+
+    return git_repo_export_dir
+
+
+def github_repo_name_format(course_id_str):
```
asadali145, line 53:
```suggestion
def github_repo_name_format(course_key_str):
```

src/ol_openedx_git_auto_export/ol_openedx_git_auto_export/utils.py
```diff
@@ -19,3 +31,71 @@ def get_publisher_username(course_module):
...
+    Format course ID to comply with GitHub repository naming conventions using slugify.
+
+    GitHub repository names:
+    - Can only contain alphanumeric characters
+    - And hyphens (-), underscores (_), and periods (.)
+    - Cannot start or end with a hyphen
+    - Maximum length is 100 characters
+
+    Args:
+        course_id_str (str): The course ID string to format
```
asadali145, line 64:
```suggestion
        course_id_str (str): The course key string to format
```

src/ol_openedx_git_auto_export/ol_openedx_git_auto_export/utils.py
```diff
@@ -19,3 +31,71 @@ def get_publisher_username(course_module):
...
+
+    Returns:
+        str: GitHub-compliant repository name
+    """
+    # Replace all characters with - hyphen except alphanumeric, hyphen, underscore, and period  # noqa: E501
+    repo_name = "".join(
+        char if char.isalnum() or char in "-_." else "-" for char in course_id_str
+    ).strip("-")
```
asadali145, line 72: Can be done with RE,
```suggestion
    repo_name = re.sub(r"[^A-Za-z0-9_.-]", "-", course_id_str).strip("-")
```

src/ol_openedx_git_auto_export/README.rst
```diff
@@ -42,8 +47,8 @@ https://help.github.com/articles/adding-a-new-ssh-key-to-your-github-account)
 
 Studio/CMS UI settings
 ----------------------
-- Open studio then course and go to advanced settings.
-- Choose field GIT URL and add you OLX git repo. For example ``git@github.com:<GITHUB_USERNAME>/edx4edxlite.git``.
+-  Open studio admin  at `/ol_openedx_git_auto_export/coursegitrepo/`
```
asadali145, line 50:
```suggestion
- Open studio admin  at `/ol_openedx_git_auto_export/coursegitrepo/`
```

asadali145 (changes requested):
In general, I have noted that we mostly use `git` instead of `GitHub`, because most of the code is specific to GitHub. Therefore, we should use GitHub in our naming conventions and documentation.

I have tested it, and it looks good.

src/ol_openedx_git_auto_export/README.rst
asadali145, line 50: We need to update this URL path.

src/ol_openedx_git_auto_export/ol_openedx_git_auto_export/tasks.py
```diff
@@ -20,14 +38,105 @@ def async_export_to_git(course_key_string, user=None):
...
+        return None
+
+    if CourseGithubRepository.objects.filter(course_id=str(course_key)).exists():
+        LOGGER.info(
+            "GitHub repository already exists for course %s. Skipping creation.",
+            course_key,
+        )
+        return None
```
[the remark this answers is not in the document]
marslanabdulrauf: In case of existing repo, it was safe to assume that github will have already have the course content.

src/ol_openedx_git_auto_export/ol_openedx_git_auto_export/utils.py
```diff
@@ -19,3 +31,71 @@ def get_publisher_username(course_module):
...
+
+    # Truncate to 100 characters if needed
+    if len(repo_name) > REPOSITORY_NAME_MAX_LENGTH:
+        repo_name = repo_name[:REPOSITORY_NAME_MAX_LENGTH].rstrip("-")
```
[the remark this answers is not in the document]
marslanabdulrauf: This is an edge case where `course_id` would cross 100 characters however I have updated it to extract last 100 characters now preserving the unique identifier (course run) so that we don't end up with duplicate repo error.

Cc @pdpinch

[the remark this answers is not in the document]
pdpinch: It's definitely an edge case that isn't worth too much time. I have a couple of suggestions: 

1. Check on the course key length limits in open edX. It's possible they can't get that long. 
2. You could remove characters at the end and replace them with a hash of the full title. That should avoid collisions. 
3. Maybe ask copilot or chatgpt.

asadali145 (approved):
LGTM!

Thanks for making all the changes.

@arslanashraf7 You can have another quick look over the changes now.

marslanabdulrauf:
I didn't receive this message may be because my local course count is less or I tested it on Enterprise.

I went through the enterprise docs for [rate limiting](https://docs.github.com/en/enterprise-server@3.4/rest/overview/resources-in-the-rest-api#rate-limiting) and found that Enterprise GitHubs do not have any rate limit. I confirmed it through [rate_limit](https://docs.github.com/en/enterprise-server@3.4/rest/rate-limit) API

PR #617, opened 2025-09-22
chore: Add Django5 support

arslanashraf7:
https://github.com/mitodl/hq/issues/7343

Adding support for Django5 while retaining support for Django4

- [ ] Desktop screenshots

### How can this be tested?
- Setup edx-platform
- Bump/Install the Django version to 5.2.6
- make sure the every plugin keeps working as expected
- Make sure that you don't see any warnings related to Django5.


asadali145 (changes requested):
I think you missed to update the Django dependency in all other packages? I see changes in edx-sysadmin only. Is this intentional?

pdpinch:
I just want to be aligned with edx-platform. Is Teak on django5, or is that not until Ulmo?

PR #590, opened 2025-08-21
fix: email matching failure during canvas grade syncs

tecoholic:
### Description (What does it do?)

In `get_student_id_by_email`, the Canvas search API does a case-insensitive search and returns the correct user object. However, the function failed to return the Canvas ID of the student as the emails capitalized in Canvas.

This commit converts both emails to lowercase during the comparison to overcome this mismatch and returns the correct user id.

### Screenshots (if appropriate):

- [ ] Desktop screenshots
- [ ] Mobile width screenshots

### How can this be tested?

Follow the testing instructions in https://github.com/mitodl/open-edx-plugins/pull/500, with 2 main differences:

1. Use the https://github.com/mitodl/edx-platform/tree/mitx/teak branch for edx-platform
2. Create the student user in Canvas with their email ALL CAPS (e.g., `LEARNER@EXAMPLE.COM`), and use the same email in all lower case for Open edX (`learner@example.com`).

* When tested with the *main* branch, the grade syncing should fail to update Canvas.
* When tested with this PR branch, the grade should sync as expected.

PR #628, opened 2025-10-06
refactor: use login_id to match canvas and openedx users

tecoholic:
Following up on #590 

When matching the users between openedx and canvas, we use the "login_id" instead of the "email" in Canvas's response, as matching on email doesn't seem to be working in production.

N/A

### How can this be tested?

Following the testing instructions in #500 

### Additional Context
- We also add a log.debug statement to log the full Canvas response. We decided to do this for short-term debugging as we don't have any visibility into Canvas's logs.

### DIFF
--- a/src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/__init__.py
+++ b/src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/__init__.py
@@ -2,6 +2,6 @@
 This is an integration of canvas with edX.
 """
 
-__version__ = "0.5.1"
+__version__ = "0.5.3"
 
 default_app_config = "ol_openedx_canvas_integration.app.CanvasIntegrationConfig"  # pylint: disable=invalid-name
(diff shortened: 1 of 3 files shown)


blarghmatey (approved):
Can you just bump the patch version as well so that the publishing goes out properly?

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/__init__.py
```diff
@@ -2,6 +2,6 @@
 This is an integration of canvas with edX.
 """
 
-__version__ = "0.5.1"
+__version__ = "0.5.3"
```
blarghmatey, line 5: The definitive location for version is in the pyproject.toml https://github.com/mitodl/open-edx-plugins/blob/main/src/ol_openedx_canvas_integration/pyproject.toml#L3

Once you update it there this is ready to merge.

blarghmatey (approved):

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/__init__.py
tecoholic: @blarghmatey I just caught the typo in pyproject.toml. Fixed it.

blarghmatey (approved):

PR #635, opened 2025-10-14
fix: standardize the tutor problem list URL setting

arslanashraf7:
None

Standardizes the tutor problem list URL setting name

- [ ] Desktop screenshots

### How can this be tested?
The tutor AI chat should work fine.


asadali145 (approved):
LGTM! We should update the package version.

arslanashraf7:
> LGTM! We should update the package version.

I explicitly didn't bump the version for this change because I wanted to take it out with any other new version after this, because this was not really an impactful change. But, I think I should update this for the sake of keeping it all standardized.

PR #642, opened 2025-10-29
feat: sync discussions configurations

asadali145:

Closes https://github.com/mitodl/hq/issues/8938

Sync discussion-related settings from source to target courses.

### How can this be tested?

- Checkout master branch
- Setup plugin following the readme. Make sure you have a source course that syncs with the target courses.
- In CMS, Open source course go to pages & resources link in Content dropdown.
- CLick on settings gear against Discussions tab.
- Now at the top right, enable the checkbox to hide the discussion tab.
- Go to the next page, change any setting or add a new topic.
- Now go to the target course and confirm that none of these changes reflects in it.
- Now checkout this branch and repeat the steps.
- Verify that all settings sync between source and target course.

### DIFF
--- a/src/ol_openedx_course_sync/ol_openedx_course_sync/utils.py
+++ b/src/ol_openedx_course_sync/ol_openedx_course_sync/utils.py
@@ -242,13 +242,13 @@ def get_syncable_course_mappings(course_key):
     Check if course sync should be performed for the given course key.
 
     Returns:
-        tuple: (should_sync: bool, course_sync_mappings: QuerySet or None)
+        Queryset: QuerySet of CourseSyncMapping or None
     """
     # Check if organization is active for sync
     if not CourseSyncOrganization.objects.filter(
         organization=course_key.org, is_active=True
     ).exists():
-        return False, None
+        return None
 
     # Check if service worker username is configured
     if not getattr(settings, "OL_OPENEDX_COURSE_SYNC_SERVICE_WORKER_USERNAME", None):
@@ -264,6 +264,6 @@ def get_syncable_course_mappings(course_key):
     )
     if not course_sync_mappings:
         log.info("No mapping found for course %s. Skipping sync.", str(course_key))
-        return False, None
+        return None
 
-    return True, course_sync_mappings
+    return course_sync_mappings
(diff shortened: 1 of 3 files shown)


src/ol_openedx_course_sync/ol_openedx_course_sync/utils.py
```diff
@@ -136,3 +149,135 @@ def update_default_tabs(source_course_key, target_course_key, user):
...
+            username=settings.OL_OPENEDX_COURSE_SYNC_SERVICE_WORKER_USERNAME
+        ).first()
+        TieredCache.set_all_tiers(cache_key, user)
+    else:
+        user = cache_value.value
+
+    return user
+
+
+def should_perform_sync(course_key):
```
arslanashraf7: How about we rename it to `get_syncable_course_runs` which returns a list. It could be an empty list by default or a populated one based on the query filter. That way, we can omit the boolean return type from here. Our check in the calling methods can be based on if a list if empty or not.

src/ol_openedx_course_sync/ol_openedx_course_sync/utils.py
```diff
@@ -136,3 +149,135 @@ def update_default_tabs(source_course_key, target_course_key, user):
...
+    )
+
+    if not has_configuration_changes:
+        return
+
+    # update discussion settings in modulestore
+    source_course = modulestore().get_course(source_course_key)
+    target_course = modulestore().get_course(target_course_key)
+
+    target_course.discussions_settings["enable_in_context"] = (
```
arslanashraf7: If target_course is None, the code below will break.

src/ol_openedx_course_sync/ol_openedx_course_sync/utils.py
```diff
@@ -136,3 +149,135 @@ def update_default_tabs(source_course_key, target_course_key, user):
...
+        source_discussions_config,
+        target_discussions_config,
+        fields_to_sync[DiscussionsConfiguration],
+    )
+
+    if not has_configuration_changes:
+        return
+
+    # update discussion settings in modulestore
+    source_course = modulestore().get_course(source_course_key)
```
arslanashraf7: We can use module_store = modulestore() and reuse that instance.

src/ol_openedx_course_sync/ol_openedx_course_sync/utils.py
asadali145: The target course can never be None because our model CourseSyncMapping already checks for that before creating a mapping.

PR #643, opened 2025-10-30
fix: replace ImportSystem with XMLImportingModuleStoreRuntime

asadali145:

None

Renames ImportSystem usages to XMLImportingModuleStoreRuntime as renamed upstream in https://github.com/openedx/edx-platform/pull/35523

### How can this be tested?

- Checks should pass
- Chat aside should work as expected.


arslanashraf7 (approved):
👍 Looks good. I noticed another issue with transcripts while testing this PR, but that's not related to this PR. I'll create a separate ticket for that.

PR #677, opened 2025-12-22
feat: add LLM course translation support

asadali145:

Closes https://github.com/mitodl/hq/issues/9664 & https://github.com/mitodl/hq/issues/9663

- Adds support to be able to translate content using different providers like DeepL, OpenAI, Gemini, and Mistral.
- Provides command line options to select the provider and the model for content and srt translations separately.
- Refactors code for better management.
- Introduces celery tasks for translations to speed up the things.
- Introduces celery task monitoring.

### How can this be tested?

- Follow the updated readme in this PR for setup and testing.

### DIFF
--- a/src/ol_openedx_course_translations/pyproject.toml
+++ b/src/ol_openedx_course_translations/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "ol-openedx-course-translations"
-version = "0.2.0"
+version = "0.3.0"
 description = "An Open edX plugin to translate courses"
 authors = [
   {name = "MIT Office of Digital Learning"}


src/ol_openedx_course_translations/ol_openedx_course_translations/management/commands/translate_course.py
```diff
@@ -46,642 +63,348 @@ def add_arguments(self, parser) -> None:
...
+            target_language = options["target_language"]
+            content_translation_provider = options["content_translation_provider"]
+            srt_provider_name = options["srt_translation_provider"]
+            glossary_directory = options.get("glossary_directory")
+            openai_model = options["openai_model"]
+            gemini_model = options["gemini_model"]
+            mistral_model = options["mistral_model"]
+
+            # Validate inputs
+            validate_course_inputs(course_archive_path)
```
arslanashraf7: We should check the provider and model validation before the course directory.

src/ol_openedx_course_translations/ol_openedx_course_translations/management/commands/translate_course.py
```diff
@@ -46,642 +63,348 @@ def add_arguments(self, parser) -> None:
...
+        )
+        parser.add_argument(
+            "--srttranslations",
+            dest="srt_translation_provider",
+            required=True,
+            choices=["deepl", "gemini", "mistral", "openai"],
+            help="AI provider to use for SRT subtitle translation.",
+        )
+        parser.add_argument(
+            "--glossary-dir",
```
arslanashraf7: Should it be a path to "--glossary-file". Because it is always going to be one glossary file being used for a single translation. That way we allow the command consumer to give any file as glossary and not make it dependent on directly and then auto guessing the file etc.

src/ol_openedx_course_translations/ol_openedx_course_translations/constants.py
arslanashraf7: I'd recommend checking all these constants here overall and deciding which could be file-level constants within their scoped files, e.g., DEEPL constants can be in deepl_provider.

src/ol_openedx_course_translations/ol_openedx_course_translations/settings/common.py
```diff
@@ -33,3 +32,14 @@ def apply_common_settings(settings):
...
+    # Translation provider API keys
+    settings.DEEPL_API_KEY = ""
+    settings.MISTRAL_API_KEY = ""
+    settings.GEMINI_API_KEY = ""
+    settings.OPENAI_API_KEY = ""
+
+    # Default model names
+    settings.OPENAI_MODEL = "gpt-5.2"
+    settings.GEMINI_MODEL = "gemini-3-pro-preview"
+    settings.MISTRAL_MODEL = "mistral-large-latest"
```
arslanashraf7: This will change to a dict as discussed over a call.

arslanashraf7 (changes requested):
This is just the code review. I'll try to test it next.

src/ol_openedx_course_translations/ol_openedx_course_translations/management/commands/translate_course.py
```diff
@@ -46,642 +63,348 @@ def add_arguments(self, parser) -> None:
         )
         parser.add_argument(
             "--course-dir",
-            dest="course_directory",
+            dest="course_archive_path",
             required=True,
             help="Specify the course directory (tar archive).",
         )
+        parser.add_argument(
+            "--contenttranslations",
```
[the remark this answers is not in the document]
asadali145: The reason behind using content is because it actually translates both the XML/HTML content and text content from files like policy.json etc. Let me know if you still think that file-translation-provider or any other name might be better here.

src/ol_openedx_course_translations/ol_openedx_course_translations/settings/common.py
```diff
@@ -54,3 +38,25 @@ def apply_common_settings(settings):
         ".xml",
         ".srt",
     ]
+    settings.DEEPL_API_KEY = ""
```
arslanashraf7: We should move this key under `TRANSLATIONS_PROVIDERS` as well. I know we won't have a default model for that but we can do with they key being empty.

arslanashraf7: Also, This note in the README will change as well `**Note:** DeepL uses the ``DEEPL_API_KEY``...`

src/ol_openedx_course_translations/ol_openedx_course_translations/management/commands/translate_course.py
```diff
@@ -46,642 +66,397 @@ def add_arguments(self, parser) -> None:
...
+            if len(parts) != 2 or not parts[0] or not parts[1]:  # noqa: PLR2004
+                error_msg = (
+                    f"Invalid provider specification: {provider_spec}. "
+                    "Use format 'PROVIDER/MODEL' (e.g., 'openai/gpt-5.2')"
+                )
+                raise CommandError(error_msg)
+            return parts[0].lower(), parts[1]
+        else:
+            # For deepl, no model is needed
+            return provider_spec.lower(), None
```
arslanashraf7: This is not just the case for DeepL, If we use srt and coursware providers without /, they also work and they also hit this case. Secondly, we can also put the default provider model logic here because it would make more sense to fetch the default model.

src/ol_openedx_course_translations/ol_openedx_course_translations/providers/base.py
```diff
@@ -0,0 +1,278 @@
...
+                logger.warning("  ... and %s more issues", len(issues) - 10)
+            return False
+        return True
+
+    def _repair_timestamps_with_deepl(
+        self,
+        original: list[srt.Subtitle],
+        target_lang: str,
+    ) -> list[srt.Subtitle]:
+        """
```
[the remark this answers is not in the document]
asadali145: @mbertrand thanks for the suggestions! do you think it would be a good idea to track such stuff in a Django model?

PR #692, opened 2026-01-14
feat: add translations persistent logs

arslanashraf7:
https://github.com/mitodl/hq/issues/9880

- Adds a new Django model to store the logs for the translated courses. The fields are
    -  source_course_id
    - source_course_language
    - target_course_language
    - srt_provider_name
    - srt_provider_model
    - content_provider_name
    - content_provider_model
    - command_stats
    - created_at
    - updated_at
- Refactors the command to print the command completion time
- Refactors the code to separateout the log messages to be used in the command and the model
- Makes the source language upper case everywhere
- Adds a new method to get the course ID from the source course tar file
- Updates some command stats

image

- [ ] Desktop screenshots

### How can this be tested?
- Setup on this branch
- Run migrations if not already
- Make sure that you are now able to see a model entry in Django Admin named `Course translation logs.`
- Now, run the same translation commands using a combination of different supported AI providers and verify that a log entry is created in the above table for each run. (You should see something like the screenshot above)


src/ol_openedx_course_translations/ol_openedx_course_translations/management/commands/translate_course.py
```diff
@@ -193,9 +195,10 @@ def _parse_and_validate_provider_spec(
     def handle(self, **options) -> None:
         """Handle the translate_course command."""
         try:
+            start_time = time.perf_counter()
             course_archive_path = Path(options["course_archive_path"])
             source_language = options["source_language"]
-            target_language = options["target_language"]
+            target_language = options["target_language"].upper()
```
asadali145: Any specific reason to convert it to upper case? In any case, shouldn't we do the same Op on source language as well?

PR #715, opened 2026-01-27
feat: add translation validation step

asadali145:

This PR adds 2 more steps when translating content.

- Validate the translated content by an LLM. Ask the LLM to rate the translation and return the issues in it.
- Ask the LLM that translated the original content to fix those issues.

### How can this be tested?

- Set up the plugin
- Translate using LLM.


asadali145:
Closing in favor of https://github.com/mitodl/open-edx-plugins/pull/718

PR #717, opened 2026-01-28
feat: remove deepl as srt backup

asadali145:

None

removes deepl as SRT fallback

### How can this be tested?

- setup the plugin and test with openai and mistral


asadali145:
@arslanashraf7 This is also ready for review, I have removed the tuning related stuff for tasks, now it only removes the DeepL repair logic.

PR #737, opened 2026-02-18
feat: library export support being added

marslanabdulrauf:
https://github.com/mitodl/hq/issues/10083

### Related `edx-platform` PR:
https://github.com/openedx/openedx-platform/pull/38026

This PR adds support to auto export libraries (v1 and v2) content to Github repositories

### How can this be tested?
1. Make sure you have set respective flags, see [Configuration](https://github.com/mitodl/open-edx-plugins/pull/737/changes#diff-20d6573c1c1ab27f196b6de26b26335141c4df4e2579b33bc8b011e8513ab582)
2. Make sure SSH is configured properly
3. Create/Update Legacy libraries
4. Check studio admin > Content Git Repositories model to see the entries
5. Check respective organization's repositories
6. Make sure you are getting latest data in latest commits
7. Repeat same steps for Library V2

### Screenshots
image

image


src/ol_openedx_git_auto_export/ol_openedx_git_auto_export/migrations/0003_migrate_course_to_content_repo.py
asadali145: As mentioned, we don't need this migration if we just rename the existing model.

src/ol_openedx_git_auto_export/ol_openedx_git_auto_export/tasks.py
```diff
@@ -25,43 +26,75 @@
...
+        # V2 libraries use content_libraries API
+        content_module = get_library(content_key)
+        content_type = "library"
+    elif is_v1_library:
+        # V1 libraries use modulestore
+        content_module = modulestore().get_library(content_key)
+        content_type = "library"
+    else:
+        content_module = modulestore().get_course(content_key)
+        content_type = "course"
```
asadali145: I think we can create a util that takes content_key and returns content_module, content_type, and url_path. It can be used in both the tasks async_export_to_git and async_create_github_repo. This will remove duplication.

asadali145 (changes requested):

src/ol_openedx_git_auto_export/ol_openedx_git_auto_export/utils.py
```diff
@@ -103,20 +147,77 @@ def export_course_to_git(course_key):
...
+
+        async_export_to_git.delay(str(library_key), user=user)
+    else:
+        log.info(
+            "Library auto-export is disabled. Skipping export for library: %s",
+            library_key,
+        )
+
+
+def is_auto_export_enabled(is_library=False):  # noqa: FBT002
```
asadali145: missing docs

src/ol_openedx_git_auto_export/ol_openedx_git_auto_export/migrations/0002_contentgitrepository.py
asadali145: We need to add another migration:
```
        migrations.AlterModelOptions(
            name='contentgitrepository',
            options={'ordering': ['-created'], 'verbose_name': 'Content Git Repository', 'verbose_name_plural': 'Content Git Repositories'},
        ),

```

asadali145 (changes requested):
We need to update the version in pyproject.toml. Also, there a couple of comments from copilot. 

Also, do we have any updates on the upstream PR because the library export won't work without those changes, and we cannot merge this PR.

marslanabdulrauf:
> Also, do we have any updates on the upstream PR because the library export won't work without those changes, and we cannot merge this PR.

yes, https://github.com/openedx/openedx-platform/pull/38026
Its also mentioned in the PR description 👍

asadali145 (approved):
LGTM!

The only thing I think we should mention in the docs is that if git export is enabled and repo creation is disabled, export won't work until we add the entry in the content repo model.

Also, we will wait for the upstream PR before we merge this one, or at least wait to enable the Library export until the upstream change is merged.

asadali145 (approved):
LGTM!

Please take a look at the couple of comments from Copilot above.

src/ol_openedx_git_auto_export/ol_openedx_git_auto_export/tasks.py
```diff
@@ -26,47 +25,70 @@
...
-            course_key_string,
+            "Git repository does not exist for %s %s. "
+            "Creating repository and exporting content.",
+            content_info["content_type"],
+            context_key_string,
         )
-        if is_auto_repo_creation_enabled():
-            async_create_github_repo.delay(str(course_key), export_course=True)
+        if is_auto_repo_creation_enabled(is_library=content_info["is_library"]):
+            async_create_github_repo.delay(str(context_key), export_content=True)
```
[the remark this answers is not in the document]
marslanabdulrauf: We don't have user information in the signal payload.
Getting it will required to extract it from `modulestore`
Follow up changes/updates in the course is passing the user detail.

If its needed we can do it in the follow-up ticket.

PR #778, opened 2026-04-01
feat: add clear expired tokens task in ol-social-auth

Anas12091101:
https://github.com/mitodl/hq/issues/10116

This PR adds a `clear_expired_tokens` task in the ol-social-auth plugin to periodically delete the expired access and refresh tokens from the edX database.

- [ ] Desktop screenshots

### How can this be tested?
- Use this [seed_tokens.py](https://github.com/user-attachments/files/26401227/seed_tokens.py) script to generate the seed data. This will create 500k expired access and refresh tokens.
- In the LMS shell run the task:
   ```python
    from ol_social_auth.tasks import clear_expired_tokens
    clear_expired_tokens.delay()
   ```
- Verify that all the expired access and refresh tokens have been deleted successfully, and non-expired tokens are still present


src/ol_social_auth/ol_social_auth/settings/common.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def plugin_settings(settings):
+    """Add clear_expired_tokens to the Celery beat schedule."""
+    settings.OAUTH2_PROVIDER["REFRESH_TOKEN_EXPIRE_SECONDS"] = (
+        30 * 24 * 60 * 60  # 30 days
+    )
+    if not hasattr(settings, "CELERY_BEAT_SCHEDULE"):
+        settings.CELERY_BEAT_SCHEDULE = {}
+    settings.CELERY_BEAT_SCHEDULE["clear_expired_tokens"] = {
```
arslanashraf7: Let's distinguish it.
```suggestion
    settings.CELERY_BEAT_SCHEDULE["ol_clear_expired_tokens"] = {
```

src/ol_social_auth/ol_social_auth/settings/common.py
```diff
@@ -0,0 +1,16 @@
...
+        "task": "ol_social_auth.tasks.clear_expired_tokens",
```
arslanashraf7: Let's do it here as well
```suggestion
        "task": "ol_social_auth.tasks.ol_clear_expired_tokens",
```

src/ol_social_auth/ol_social_auth/settings/common.py
```diff
@@ -0,0 +1,16 @@
...
+        "schedule": crontab(hour=9, minute=0, day_of_week="monday"),
```
arslanashraf7, line 16: Why not midnight?
```suggestion
        "schedule": crontab(hour=0, minute=0, day_of_week="monday"),
```

src/ol_social_auth/ol_social_auth/settings/common.py
```diff
@@ -0,0 +1,16 @@
+"""Common settings for the ol-social-auth plugin."""
+
+from celery.schedules import crontab
+
+
+def plugin_settings(settings):
+    """Add clear_expired_tokens to the Celery beat schedule."""
```
arslanashraf7: The docstrings are for the while plugin_settings, so they need to be generic.

src/ol_social_auth/ol_social_auth/tasks.py
```diff
@@ -0,0 +1,25 @@
...
+oauth2_logger = logging.getLogger("oauth2_provider")
+
+
+@shared_task(acks_late=True)
+def clear_expired_tokens():
+    """Clear expired OAuth2 access, refresh, and ID tokens."""
+    log.info("Starting clear_expired_tokens...")
+    # Suppress debug-level logs from oauth2_provider during cleanup.
+    # Its batch_delete debug logs lack the 'userid' field expected by
+    # Open edX's custom log formatter, causing noisy ValueError tracebacks.
```
arslanashraf7, line 18: Can you explain this a little bit more? I am having a hard time getting it. Did we copy it from Open edX?

src/ol_social_auth/README.rst
```diff
@@ -32,3 +32,15 @@ Make sure to properly configure the plugin following the links in the above "Con
...
 * Verify that you are not logged in on edx-platform.
 * Create a new user in your MIT application and verify that a corresponding user is successfully created on the edX platform.
+
+Expired Token Cleanup
+---------------------
+This plugin includes a scheduled Celery task (``clear_expired_tokens``) that automatically removes expired OAuth2 access tokens, refresh tokens, and grant tokens from the database.
+
+**Behavior:**
+
+* Runs every **Monday at 9:00 AM** (server time) via Celery Beat by default. The schedule can be customized by overriding the ``clear_expired_tokens`` entry in ``CELERY_BEAT_SCHEDULE``.
```
arslanashraf7: this will change too if we decide to run the task at midnight.

arslanashraf7 (changes requested):

src/ol_social_auth/ol_social_auth/settings/common.py
Anas12091101: I’m not entirely sure, but this task is scheduled for 9 AM in our other applications as well; for example, [MITxOnline](https://github.com/mitodl/mitxonline/blob/main/main/settings.py#L1022) and [MITxPRO](https://github.com/mitodl/mitxpro/blob/pre-commit-ci-update-config/mitxpro/settings.py#L849-L852)

@pdpinch, what are your thoughts on this?

pdpinch: If it's a task that we care about succeeding, I prefer to do it during business hours, so we have a better chance of noticing a problem and fixing it. 

For this task, like this, really, any time would be fine.

src/ol_social_auth/ol_social_auth/tasks.py
Anas12091101: This wasn't copied from Open edX — it's a workaround for an incompatibility between `django-oauth-toolkit` and Open edX's custom log formatter.

**The problem:** `clear_expired()` internally calls `batch_delete()`, which emits debug-level log messages like `"10000 tokens deleted, X left"`. Open edX configures a custom log formatter that expects every log record to have a `userid` field (used for request-scoped user tracking). The `oauth2_provider` library doesn't add this field to its log records, so every debug message it emits causes a `ValueError: Formatting field not found in record: 'userid'` — producing a noisy traceback in the logs on every batch deletion.

**The fix:** Before calling `clear_expired()`, we temporarily raise the `oauth2_provider` logger level to `INFO`, which silences the debug-level batch messages. After the call completes (or fails), we restore the original level. This way the deletion still happens correctly — we just suppress the log messages that would trigger the formatter error.

To reproduce the issue, run the clear_expired() fn without suppressing the logs.

src/ol_social_auth/ol_social_auth/settings/common.py
```diff
@@ -0,0 +1,17 @@
...
+    """Settings for the ol-social-auth plugin."""  # noqa: D401
+    settings.OAUTH2_PROVIDER["REFRESH_TOKEN_EXPIRE_SECONDS"] = (
+        30 * 24 * 60 * 60  # 30 days
+    )
+    # Add ol_clear_expired_tokens to the Celery beat schedule.
+    if not hasattr(settings, "CELERY_BEAT_SCHEDULE"):
+        settings.CELERY_BEAT_SCHEDULE = {}
+    settings.CELERY_BEAT_SCHEDULE["ol_clear_expired_tokens"] = {
+        "task": "ol_social_auth.tasks.ol_clear_expired_tokens",
+        "schedule": crontab(hour=9, minute=0, day_of_week="monday"),
```
arslanashraf7: I was able to test it manually via Django shell, but I wasn't able to test it via the beat schedule and the task being run in auto mode. Were you able to test it?

Also, Open edX has `CELERYBEAT_SCHEDULE` in the settings in [common.py](https://github.com/openedx/openedx-platform/blob/master/lms/envs/common.py#L1713) instead of `CELERY_BEAT_SCHEDULE`. Could you tell if there was any specific reason for using a new env key rather than using the existing openedx one? And also, could this be the reason the task schedule isn't working?

arslanashraf7 (changes requested):

src/ol_social_auth/ol_social_auth/settings/common.py
Anas12091101: Good catch. I’ve updated the setting name in https://github.com/mitodl/open-edx-plugins/pull/778/commits/dce4d4b8274a7ee75994576c2ffc4957e4d5a027 and verified that the tasks are running as expected according to the beat schedule.

src/ol_social_auth/ol_social_auth/settings/common.py
```diff
@@ -0,0 +1,17 @@
+"""Common settings for the ol-social-auth plugin."""
+
+from celery.schedules import crontab
+
+
+def plugin_settings(settings):
+    """Settings for the ol-social-auth plugin."""  # noqa: D401
+    settings.OAUTH2_PROVIDER["REFRESH_TOKEN_EXPIRE_SECONDS"] = (
+        30 * 24 * 60 * 60  # 30 days
```
arslanashraf7, line 9: I'd suggest checking with @blarghmatey this time range as well, since we're using service tokens before merging the PR. Since DevOps creates service tokens, I'm not sure what expiry time they set. I want to make sure they will not be deleted automatically.

arslanashraf7 (approved):
LGTM 👍 One question for @blarghmatey before merging.

src/ol_social_auth/ol_social_auth/settings/common.py
blarghmatey: The OAuth service accounts that we rely on are using client credentials grants so we're not worried about token expiration. Thanks for checking.

PR #788, opened 2026-05-08
feat: add Canvas due date sync functionality

xitij2000:
N/A

This change introduces functionality to synchronize due dates between Open edX and Canvas within the `ol_openedx_canvas_integration` plugin. It allows Canvas to act as the source of truth for assignment due dates.

Key features include:
- **New Course Setting**: Added `use_canvas_due_dates` to course settings to toggle this behavior.
- **Assignment Sync Logic**: When enabled, pushing grades or assignments to Canvas will no longer overwrite Canvas due dates with Open edX dates.
- **Canvas to Open edX Sync**: A new Celery task `sync_canvas_due_dates` fetches `due_at` dates from Canvas assignments and updates the corresponding subsection blocks in Open edX.
- **Management Command**: Added `python manage.py sync_canvas_due_dates` to allow manual triggering of the due date sync for specific courses or globally.
- **Improved Testing**: Updated the integration test runner to support CMS-based tests for this plugin.

### How can this be tested?
1. **Enable the Setting**: In Open edX course settings (Other Course Settings), set `"use_canvas_due_dates": true`.
2. **Push to Canvas**: Verify that when assignments are created or updated in Canvas via the plugin, the due dates in Canvas remain unchanged (or are not sent if being created).
3. **Sync from Canvas**:
   - Change a due date for an assignment in Canvas.
   - Run the management command: `python manage.py sync_canvas_due_dates <course_id>`.
   - Verify in Studio/LMS that the subsection's due date has been updated to match Canvas.
4. **Automated Tests**:
   - Run the plugin tests using the integration script: `./run_edx_integration_tests.sh --plugin ol_openedx_canvas_integration`.
   - Specifically check `src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/test_cms_tasks.py` for coverage of the sync logic.

### Additional Context
This change requires a mechanism to run management command on schedule. If there are currently no other scheduled tasks, then this might require installing a plugin. [`tutor-contrib-grove`](https://gitlab.com/opencraft/dev/tutor-contrib-grove/-/tree/main/tutorgrove/plugins?ref_type=heads) for instance support creating cron jobs. 

private-ref: https://tasks.opencraft.com/browse/BB-10752


src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/management/commands/sync_canvas_due_dates.py
```diff
@@ -0,0 +1,89 @@
...
+                    f"Processing all {courses.count()} course(s) in the system..."
+                )
+            )
+        for course_id in courses:
+            try:
+                sync_canvas_due_dates.delay(str(course_id))
+            except Exception as ex:
+                self.stderr.write(
+                    self.style.ERROR(f"Error processing course {course_id}: {str(ex)}")
+                )
```
tecoholic, line 105: How often does this error occur? Since things are being queued with "delay", the errors should happen only in the background worker right?

tecoholic: Oh! wait a sec, tutor dev runs celery in "immediate mode". So errors bubble up..

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/cms_tasks.py
```diff
@@ -176,3 +183,54 @@ def sync_course_assignments_with_canvas(course_id):
...
     update_assignments(canvas, operations_map["update"])
     delete_assignments(canvas, operations_map["delete"])
+
+
+@shared_task
+def sync_canvas_due_dates(course_id: str):
+    _sync_canvas_due_dates(course_id)
+
+
+def _sync_canvas_due_dates(course_id: str):
```
tecoholic, line 273: @xitij2000 Kindly add a docstring for this function.

run_edx_integration_tests.sh
```diff
@@ -179,7 +179,7 @@ run_plugin_tests() {
     fi
 
     # Run the pytest command with CMS settings (for ol_openedx_chat)
```
tecoholic: I think we can update the comment to reflect the current state
```suggestion
    # Run the pytest command with CMS settings for plugins with CMS specific functionality
```

src/ol_openedx_canvas_integration/tests/test_cms_tasks.py
```diff
@@ -0,0 +1,210 @@
...
+        with (
+            patch("ol_openedx_canvas_integration.cms_tasks.TASK_LOG") as mocked_log,
+        ):
+            _sync_canvas_due_dates(str(course.id))
+
+            mocked_log.info.assert_called_with(
+                "Due Date Sync: No canvas ID. Skipped for course %s",
+                str(course.id),
+            )
+
```
tecoholic, line 170: Verifying Canvas API calls are not made would be a better assertion of the functionality than a logging statement. This turns into a pain latter if the statement is changed, other log statements or added..etc.,

src/ol_openedx_canvas_integration/tests/test_cms_tasks.py
```diff
@@ -0,0 +1,210 @@
...
+        course, _ = self.create_course({
+            "canvas_id": 11,
+            **other_course_settings,
+        })
+        with (
+            patch("ol_openedx_canvas_integration.cms_tasks.TASK_LOG") as mocked_log,
+        ):
+            _sync_canvas_due_dates(str(course.id))
+
+            mocked_log.info.assert_called_with(
```
tecoholic: Same comment as above.

tecoholic (changes requested):
@xitij2000 Hi, this is pretty close to done. My canvas installation is out of order. I will set it up and test the flow fully. Meanwhile, kindly take a look at the comments.

Nice work setting up the testing infrastructure. I learnt about the cms specific utilities through this PR. Thank you for that.

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/management/commands/sync_canvas_due_dates.py
xitij2000: Honestly, I added this because I initially did this without celery since that's easier to test. Either way if any error can reach here, it's better it get logged and to break other courses from processing.

src/ol_openedx_canvas_integration/tests/test_cms_tasks.py
xitij2000: Certainly!

tecoholic:
@xitij2000 The instructor dashboard on tutor-main didn't work for me. It seems to have been caused by [this PR introducing instructor.legacy_instructor_dashboard](https://github.com/openedx/openedx-platform/issues/38432) Waffle flag.

However, after setting the waffle flag, I couldn't get any of the buttons to work on the legacy dashboard to trigger the actual sync to go through. 

[Screencast_20260511_220052.webm](https://github.com/user-attachments/assets/ba78a69e-7b95-4239-8242-8fe719fe148f)

tecoholic:
@xitij2000 I tried to switch to a Ulmo release based devstack to avoid the instructor-dashboard issue. It didn't work either

image

It looks like the Canvas integration on Ulmo might be broken?

xitij2000:
@tecoholic I just directly ran the management commant via `tutor dev exec ...`

tecoholic:
@xitij2000 Ah! I thought we need to do the initial sync from the UI to get the Open edX assignments appear on the Canvas course and then the managemnent command we introduce in this PR, keeps the dates in sync on the Open edX end. 

It's pretty late here, so maybe I am not thinking straight. I will test it again tomorrow.

tecoholic:
@xitij2000 :+1:  I was able to test this. I have completely overlooked the fact that the assingments are actually automatically synced on "XBLOCK_PUBLISHED" signals. All I needed to do was just add an extra block and hit publish. The due date syncing works as expected.

* Can you kindly bump the version in "pyproject.toml"?
* Add the feature to the README?
* Fix the CI failure. `ruff format....................................Failed`

:+1:

- [x] I tested this: verified that the command syncs the assignment due dates and canvas dates are not overwritten when changed in Open edX.
- [x] I read through the code
- [x] I checked for accessibility issues
- [ ] Includes documentation - requested.

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/tasks.py
```diff
@@ -110,6 +119,16 @@ def sync_user_grade_with_canvas(grade_id):
         )
         return
 
+    due_date = parse_datetime(
+        existing_assignments_map[str(grade_instance.usage_key)]["due_at"]
+    )
+    if due_date and due_date < datetime.now():
```
tecoholic: @xitij2000 Grade syncing fails here with the following errors:

```
lms-1  | 2026-05-21 06:18:41,782 ERROR 53 [celery.app.trace] [user 5] [ip 172.24.0.1] trace.py:309 - Task ol_openedx_canvas_integration.tasks.sync_user_grade_with_canvas[cb184043-eb80-4c9d-85c1-593a2a34d9d0] raised unexpected: TypeError("can't compare offset-naive and offset-aware datetimes")
lms-1  | Traceback (most recent call last):
lms-1  |   File "/openedx/venv/lib/python3.12/site-packages/celery/app/trace.py", line 585, in trace_task
lms-1  |     R = retval = fun(*args, **kwargs)
lms-1  |                  ^^^^^^^^^^^^^^^^^^^^
lms-1  |   File "/openedx/venv/lib/python3.12/site-packages/ol_openedx_canvas_integration/tasks.py", line 92, in sync_user_grade_with_canvas
lms-1  |     _sync_user_grade_with_canvas(grade_id)
lms-1  |   File "/openedx/venv/lib/python3.12/site-packages/ol_openedx_canvas_integration/tasks.py", line 125, in _sync_user_grade_with_canvas
lms-1  |     if due_date and due_date < datetime.now():
lms-1  |                     ^^^^^^^^^^^^^^^^^^^^^^^^^
lms-1  | TypeError: can't compare offset-naive and offset-aware datetimes
```

xitij2000: Surprising that I didn't get this error. Will fix.

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/cms_tasks.py
```diff
@@ -176,3 +200,119 @@ def sync_course_assignments_with_canvas(course_id):
...
+                    student.id,
+                    course.id,
+                )
+                set_due_date_extension(
+                    course,
+                    block,
+                    student,
+                    override["due_at"],
+                    reason=f"Synced from canvas course: {canvas_course_id}",
+                )
```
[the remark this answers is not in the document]
xitij2000: This does work without conversion since the Django does it implicitly later down the line. However, an explicit conversion here shouldn't hurt.

arslanashraf7:
@xitij2000 could you fix the failing pre-commit check?

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/management/commands/sync_canvas_due_dates.py
```diff
@@ -0,0 +1,87 @@
+"""
+Script for syncing Canvas due dates with Open edX
```
arslanashraf7, line 2: nit: It would be nice to add docs on sample runs of this command.

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/utils.py
```diff
@@ -6,6 +6,15 @@ def get_canvas_course_id(course=None):
     return course.other_course_settings.get("canvas_id") if course else None
 
 
+def get_use_canvas_due_dates(course=None):
+    """Get the canvas due dates setting from the course settings"""
+    return (
+        course.other_course_settings.get("use_canvas_due_dates", False)
+        if course
+        else None
```
arslanashraf7: Let's keep the response consistent.

```suggestion
        else False
```

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/management/commands/sync_canvas_due_dates.py
```diff
@@ -0,0 +1,87 @@
...
+            )
+        else:
+            self.stdout.write(
+                self.style.SUCCESS(
+                    f"Processing all {courses.count()} course(s) in the system..."
+                )
+            )
+        for course_id in courses:
+            try:
+                sync_canvas_due_dates.delay(str(course_id))
```
arslanashraf7, line 101: Should we also check whether `get_use_canvas_due_dates` is enabled ? Skip if disabled?

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/utils.py
```diff
@@ -6,6 +6,15 @@ def get_canvas_course_id(course=None):
     return course.other_course_settings.get("canvas_id") if course else None
 
 
+def get_use_canvas_due_dates(course=None):
```
arslanashraf7: This might be better for readability.
```suggestion
def is_canvas_dates_sync_enabled(course=None):
```

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/client.py
```diff
@@ -246,32 +268,39 @@ def update_assignment_grades(self, canvas_assignment_id, payload):
...
+            if not subsection_block.fields.get("due")
+            # The internal API gives us a TZ-naive datetime for the due date, but Studio indicates that  # noqa: E501
+            # the user should enter a UTC datetime for the due date. Coerce this to UTC before creating the  # noqa: E501
+            # string representation.
+            else subsection_block.fields["due"].astimezone(pytz.UTC).isoformat()
+        ),
+    }
+    # If we're using canvas dates, don't set the due dates from Open edX
+    if use_canvas_dates:
+        due_date_dict = {}
```
arslanashraf7: Should this be done in an if/else? or something like:

```
due_date_dict = {}

if not use_canvas_dates:
    ... (due date calculation code)

File: src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/client.py [hunk 13]
```diff
@@ -289,4 +318,4 @@ def update_grade_payload_kv(user_id, grade_percent):
     Returns:
         (tuple): A key/value pair that will be used in the body of a bulk grade update request
     """  # noqa: D401, E501
-    return (f"grade_data[{user_id}][posted_grade]", f"{grade_percent * 100}%")
+    return f"grade_data[{user_id}][posted_grade]", f"{grade_percent * 100}%"
```
arslanashraf7, line 321: Was there a specific reason for this change? If so, The Return value in docstrings should also be updated because it won't return a tuple anymore.

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/cms_tasks.py
```diff
@@ -176,3 +200,120 @@ def sync_course_assignments_with_canvas(course_id):
...
+                usage_key = UsageKey.from_string(usage_id)
+                due_at = canvas_assignment.get("due_at")
+                block = modulestore().get_item(usage_key)
+                sync_canvas_due_date_extensions(
+                    client, course, block, canvas_assignment.get("overrides")
+                )
+                if due_at:
+                    block.due = parse_datetime(due_at)
+                else:
+                    block.due = None
```
arslanashraf7, line 323: If the use of Canvas due dates is true but there is no due date from Canvas, should we fall back to the existing due date set in Open edX, if any, or force it to `None`?

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/tasks.py
```diff
@@ -110,6 +119,16 @@ def sync_user_grade_with_canvas(grade_id):
...
 
+    due_date = existing_assignments_map[str(grade_instance.usage_key)]["due_at"]
+    if due_date:
+        due_date = parse_datetime(due_date)
+    if due_date and due_date < datetime.now(tz=UTC):
+        TASK_LOG.warning(
+            "The assignment %s is past its due date. Skipping grade sync.",
+            grade_instance.usage_key,
+        )
+        return
```
arslanashraf7, line 130: This is done in your other PR. Should we remove it from here?

src/ol_openedx_canvas_integration/tests/conftest.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+def pytest_configure():
+    """Pytest hook that runs after command line options have been parsed"""
+
+    # Add additional Django settings needed for the plugin tests
+    if not hasattr(settings, "BULK_EMAIL_DEFAULT_RETRY_DELAY"):
+        settings.BULK_EMAIL_DEFAULT_RETRY_DELAY = 10
+    if not hasattr(settings, "BULK_EMAIL_MAX_RETRIES"):
+        settings.BULK_EMAIL_MAX_RETRIES = 5
```
arslanashraf7, line 15: Likewise, this is the code from your other PR, but i suppose you needed it here to get the tests working. Right?

src/ol_openedx_canvas_integration/tests/test_cms_tasks.py
```diff
@@ -0,0 +1,283 @@
...
+
+            for seq_id, data in mock_canvas_assignments.items():
+                seq_key = UsageKey.from_string(seq_id)
+                due_at = data.get("due_at", None)
+                due_at = parse_datetime(due_at) if due_at else None
+                assert self.store.get_item(seq_key).due == due_at
+
+    def test_sync_canvas_due_date_extensions(self):
+        for uid in [1, 4, 9, 11, 14, 37]:
+            User.objects.create_user(f"user{uid}", f"user{uid}@abc.xyz", "password")
```
arslanashraf7: nit: Should we use UserFactory from edX instead?

src/ol_openedx_canvas_integration/tests/test_tasks.py
```diff
@@ -0,0 +1,236 @@
...
+    @patch("ol_openedx_canvas_integration.tasks.CanvasClient")
+    @patch("ol_openedx_canvas_integration.tasks.get_subsection_user_grades")
+    @patch("ol_openedx_canvas_integration.tasks.TASK_LOG")
+    @patch(
+        "ol_openedx_canvas_integration.tasks.get_canvas_course_id",
+        MagicMock(return_value="canvas-123"),
+    )
+    def test_sync_success(
+        self,
+        mock_task_log,
```
arslanashraf7, line 68: This and some other tests are also from your other PR. My assumption is that we'll need to first merge that, and then this will be rebased.

src/ol_openedx_canvas_integration/README.rst
```diff
@@ -81,7 +81,8 @@ Configuration
 1) Open your course in Studio.
 2) Navigate to "Advanced Settings".
 3) Enable other course settings by enabling ``ENABLE_OTHER_COURSE_SETTINGS`` feature flag in CMS
-4) Open course advanced settings in Open edX CMS, Add a dictionary in ``{"canvas_id": <canvas_course_id>}``. The ``canvas_course_id`` should be the id of a course that exists on Canvas. (NOTE: Canvas tab would only be visible if this value is set)
+4) Open course advanced settings in Open edX CMS, Add a dictionary in ``{"canvas_id": <canvas_course_id>, "use_canvas_due_dates": <True/False>}``. The ``canvas_course_id`` should be the id of a course that exists on Canvas.
+   The ``use_canvas_due_dates`` is an optional flag (defaults to ``False``) to sync the assignment due dates from Canvas to Open edX. (NOTE: Canvas tab would only be visible if ``canvas_id`` is set)
```
arslanashraf7, line 85: nit: Maybe we can do this too since we are updating it:


```suggestion
4) Open course advanced settings in Open edX CMS. Add a dictionary in ``{"canvas_id": <canvas_course_id>, "use_canvas_due_dates": <True/False>}``. The ``canvas_course_id`` should be the id of a course from Canvas.
   The ``use_canvas_due_dates`` is an optional flag (defaults to ``False``) to sync the assignment due dates from Canvas to Open edX. (NOTE: Canvas tab would only be visible if ``canvas_id`` is set)
```

src/ol_openedx_canvas_integration/README.rst
```diff
@@ -125,3 +127,31 @@ Whenever the course is **Published** from the Studio, the **graded subsections**
...
 Whenever a learner interacts with a graded question in Open edX, the latest grades are automatically posted to Canvas, if it's a part of a synced assignment.
+
+3. Automatic Syncing of Due Dates
+""""""""""""""""""""""""""""""
+
+This feature allows for periodic syncing of assignment due dates from the linked Canvas course to Open edX.
+
+**Scheduling the Task**
+
+This sync does not happen automatically. Instead, it is triggered by a management command that needs to be scheduled as a periodic task (e.g., via cron).
```
arslanashraf7: Is there a specific reason for scheduling the management command and not the tasks themselves? Or how about adding a new button in the Canvas tab in the instructor dashboard to sync the dates from Canvas to Open edX?

cc: @pdpinch

run_edx_integration_tests.sh
```diff
@@ -183,7 +183,7 @@ run_plugin_tests() {
...
```
arslanashraf7: Need to update this comment as well.

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/management/commands/sync_canvas_due_dates.py
```diff
@@ -0,0 +1,87 @@
...
+
+        if not course_keys and not all_courses:
+            # Either course_keys or --all flag must be provided
+            command_name = Path(__file__).stem
+            self.stderr.write(
+                self.style.ERROR(
+                    "Error: You must specify either course keys or use the --all flag.\n"  # noqa: E501
+                    "Examples:\n"
+                    f"  python manage.py {command_name} --all\n"
+                    f"  python manage.py {command_name} course-v1:MITx+6.00x+2T2024"
```
arslanashraf7: Since these are meant to run in CMS, they should be rather specific 
```suggestion
                    f"  python manage.py cms {command_name} --all\n"
                    f"  python manage.py cms {command_name} course-v1:MITx+6.00x+2T2024"
```

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/management/commands/sync_canvas_due_dates.py
```diff
@@ -0,0 +1,87 @@
...
+            )
+            return
+
+        courses = (
+            CourseOverview.objects.all()
+            .order_by("created")
+            .values_list("id", flat=True)
+        )
+        if course_keys:
+            courses = courses.filter(id__in=course_keys)
```
arslanashraf7, line 87: This could throw `opaque_keys.InvalidKeyError` if an invalid course key is specified. We should show a more descriptive command error instead of default raw error stacktrace.

arslanashraf7 (changes requested):
The functionality looks good. Suggested some changes and added some questions, though.

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/client.py
xitij2000: Brackets are not required here and some linting tools throw an error, simply returning two values as `a, b` will work.

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/management/commands/sync_canvas_due_dates.py
xitij2000: Done!

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/management/commands/sync_canvas_due_dates.py
xitij2000: The task will already do that. On order to know if that option is enabled, the course will needed to be loaded from mongodb twice, once here, and then once in the celery task, so while it will save the celery overhead for tasks that can be skipped, it will add a double load overhead for tasks that won't be skipped.

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/utils.py
xitij2000: While I think False and None convey different information, i.e. False is specificlly about the setting being False, None conveys that the course itself doesn't exist. However, I can't imagine this being used independently so the extra info is moot.

src/ol_openedx_canvas_integration/tests/conftest.py
xitij2000: I think it might be better to get this merged first, and then I'll rebase and remove any redundancies in the other PR.

src/ol_openedx_canvas_integration/tests/test_cms_tasks.py
xitij2000: Done!

src/ol_openedx_canvas_integration/tests/test_tasks.py
xitij2000: Yes! Since the tests were not running before this PR, I had to fix some tests, and those fixes need to be in both PRs to get tests to pass. After merging this one I'll rebase and fix that PR as you mentioned.

src/ol_openedx_canvas_integration/README.rst
xitij2000: The idea is for this to be automatic without requiring any manual syncing. That was part of the scope of this task. To schedule the task itself we'd need  celerybeat configured. If that is configured, I can simply add a task to run this for all courses.

arslanashraf7: The Celerybeat I think is configured. We have some scheduled tasks in another plugin that we recently built https://github.com/mitodl/open-edx-plugins/tree/main/src/ol_openedx_events_handler.

@blarghmatey is Celerybeat configured in MITx Residential? And what is your take on this between a scheduled Celery task vs. a scheduled command via tutor-contrib-groove?

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/cms_tasks.py
```diff
@@ -176,3 +200,120 @@ def sync_course_assignments_with_canvas(course_id):
...
+            emails = client.get_emails_by_student_ids(override["student_ids"])
+            students = User.objects.filter(email__in=emails)
+            for student in students:
+                TASK_LOG.info(
+                    "Due Date Sync: Syncing due date for student %s in course %s",
+                    student.id,
+                    course.id,
+                )
+                due_date_override = parse_datetime(override["due_at"])
+                set_due_date_extension(
```
arslanashraf7, line 264: Assume this case where this might fail:

The instructor sets a due date for a student in Canvas; the date gets synced with edX, and then later on the instructor clears that date in Canvas. It will never be updated back in Open edX, and the due date for the user will stay the same as what it was set to earlier.

But this is an edge case we can't really reliably solve. An option is to use `dump_student_extensions` from Open edX, but the problem is when to call it and for which blocks to call it so that it only resets the dates for the block dates that were only overridden by the Canvas.

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/cms_tasks.py
```diff
@@ -176,3 +200,120 @@ def sync_course_assignments_with_canvas(course_id):
...
+    the platform.
+
+    Parameters:
+        course_id (str): The unique identifier of the course whose due
+        dates need to be synchronized.
+    """
+    _sync_canvas_due_dates(course_id)
+
+
+def sync_canvas_due_date_extensions(client, course, block, overrides):
```
arslanashraf7, line 240: A problem that I'm seeing here is that when you set the due date for a student in Canvas and sync the course, the Admins see that due date in CMS/Studio as well, which should not be the case.

If you set an extension via `lms/Instructor dashboard/extensions` it works fine and the due date is only updated for that student in LMS.

arslanashraf7 (changes requested):

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/cms_tasks.py
xitij2000: I think one solution to this would be to advise canvas authors to not delete the extension, but to set it to the non-extended date, or a minute or two after it.

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/cms_tasks.py
xitij2000: I'm not sure I understand. Are you saying that the extended due date is applied for all students in addition to creating an extension?

arslanashraf7: This is the scenario:

**Step 1:** Without this plugin, set an extension for a student in edX via the Instructor Dashboard, visit the LMS with that student signed in, and see the due date; you will see the extended date as expected.
**Step 2:** Now visit CMS, open the course outline, and see the due date and you see the overall due date for the subsection that's for all the other students. This also works as expected.

Now,

- Run this plugin
- Set an extension date for a student in Canvas
- Sync the due dates via the command (`./manage.py cms sync_canvas_due_dates --all`) in this PR 
- Perform step 1 from above, and the result is as expected -> You see the extended date we set for that student in Canvas
- Now perform Step 2. Here is the problem: the course author sees the due date that was set for a single student in Canvas instead of the date that was set for all the other students overall.

Let me know if this is still confusing, and I will share a screen recording with you.

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/cms_tasks.py
arslanashraf7: Yes, I agree; that might be a suitable option here.

src/ol_openedx_canvas_integration/README.rst
blarghmatey: We don't use Tutor at all, so that's not helpful for us, but doesn't hurt to have available if other folks might want to use this. We do have celerybeat and should use that since it's well-understood machinery.

xitij2000: Sure, I'll add a celerybeat config for this.

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/cms_tasks.py
xitij2000: Okay, I understand the issue, and am looking into it.

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/cms_tasks.py
xitij2000: I think for consistency it should be set to `None`, since otherwise there is no way to unset a data once sent. I think if there is need for more complex syncing scenarios, then we'll probably need a intermediary syncing app with a UI to allow selecting what to do in each situation.

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/cms_tasks.py
xitij2000: I think I accidentally left out a param for this. Basically `override_assignment_dates` is set to true by default. When true it will return the due date that applies to the user whose access token is being used. I've now set that to false so it will return the correct due date.

src/ol_openedx_canvas_integration/README.rst
xitij2000: I've added the celery-beat config.

arslanashraf7:
@xitij2000 let me know when this is ready for another look.

xitij2000:
> @xitij2000 let me know when this is ready for another look.

@arslanashraf7  I think it's ready from the code POV, but I'm trying to figure out how to manually test it with celery beat.

xitij2000:
@arslanashraf7 It's ready for another review pass.

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/settings/cms/common.py
arslanashraf7, line 1: nit: We aren't maintaining LMS/CMS specific separate settings for our plugins. Could you move your code to the existing common.py and production.py files in this plugin?

arslanashraf7 (commented):
@xitij2000 This looks good, and this worked well in testing as well. Just one comment i need your thoughts about.

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/settings/cms/common.py
xitij2000: We don't want the celery beat tasks to be running with LMS settings since it requires CMS settings. That is why this needed a split. Also it will require a new celery beat running with cms settings if it isn't already

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/cms_tasks.py
```diff
@@ -176,3 +201,138 @@ def sync_course_assignments_with_canvas(course_id):
     add_assignments(canvas, operations_map["add"])
     update_assignments(canvas, operations_map["update"])
     delete_assignments(canvas, operations_map["delete"])
+
+
+@shared_task
+def sync_canvas_due_dates_for_all_courses():
```
arslanashraf7, line 207: Can we add a test for this as well?

arslanashraf7 (commented):
@xitij2000 Just last request before this can be approved and merged.

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/cms_tasks.py
xitij2000: Added

arslanashraf7: @xitij2000 I've run the CI workflow on your PR. Could you take a look at the failing pre-commit as well?

xitij2000: I've applied the fixes.

arslanashraf7 (approved):
👍 Thanks for making all the changes.

arslanashraf7:
@xitij2000 I've merged this PR. You'll now need to rebase https://github.com/mitodl/open-edx-plugins/pull/804. Let me know done, and I'll give that a final go.

PR #791, opened 2026-05-11
config: add ol_clear_expired_tokens task in production settings

Anas12091101:
https://github.com/mitodl/hq/issues/10116

The task is not appearing in the `CELERYBEAT_SCHEDULE` on RC, despite `ol-social-auth` 0.2.1 being installed. This PR adds the `ol_clear_expired_tokens` task in production.py. I am not entirely sure how the deployment flow works, but according to the co-pilot, this is the ordering:

1. The plugin's common.py `plugin_settings()` runs during common.py loading (early) — this correctly adds `ol_clear_expired_tokens` with a real `crontab()` object.
2. The YAML config (loaded in production.py ~line 70) overwrites `CELERYBEAT_SCHEDULE` entirely via `vars().update()`, wiping out what common.py set.
3. `refresh-saml-metadata` is added back by the platform's own production.py at line 322 (hardcoded post-YAML logic) — which is why it still appears in RC.
4. `add_plugins(__name__, ProjectType.LMS, SettingsType.PRODUCTION)` runs at line 441 — **after** the YAML override — so anything set in a plugin's production.py survives.

By duplicating the schedule entry in production.py, it gets re-added after the YAML wipe and persists at runtime. The common.py entry is kept as-is so local dev environments continue to work.

- [ ] Desktop screenshots

### DIFF
--- a/src/ol_social_auth/ol_social_auth/settings/production.py
+++ b/src/ol_social_auth/ol_social_auth/settings/production.py
@@ -4,7 +4,7 @@ from celery.schedules import crontab
 
 
 def plugin_settings(settings):
-    """Production overrides for ol-social-auth plugin."""  # noqa: D401
+    """Production overrides for ol-social-auth plugin."""
     # Re-add the celery beat schedule here because the YAML config loading
     # in production.py replaces the entire CELERYBEAT_SCHEDULE dict,
     # wiping out what common.py set. This runs after YAML loading.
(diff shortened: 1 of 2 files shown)


arslanashraf7 (approved):
👍 LGTM

You'll need to bump the version in `pyproject.toml` before merging this PR.

PR #796, opened 2026-05-18
fix: wrap-instructor-info css to correctly display staff buttons on Show Answer

Anas12091101:
https://github.com/mitodl/hq/issues/11304

When a learner clicks "Show Answer" on a problem, the revealed answer content expands within a flex parent container, causing .wrap-instructor-info (Staff Debug Info / Submission History buttons) to collapse to zero width. This adds width: 100%, flex-shrink: 0, and flex-wrap: wrap to ensure the buttons maintain their layout regardless of sibling content expansion.

Before
Screenshot 2026-05-18 at 5 33 28 PM

After
Screenshot 2026-05-18 at 5 31 38 PM

### How can this be tested?
- Enabling the AskTim button is not required for testing. The issue can be reproduced without it.
- Install ol-openedx-chat from this branch.
- Enable the theme, and open a problem.
- Click the "Show answer" button.
- Verify that it doesn't break the styling of Staff Debug Info / Submission History buttons.

### DIFF
--- a/src/ol_openedx_chat/ol_openedx_chat/static/css/ai_chat.css
+++ b/src/ol_openedx_chat/ol_openedx_chat/static/css/ai_chat.css
@@ -113,3 +113,13 @@
 .video {
     margin: 0 0;
 }
+
+@media (min-width: 768px) {
+  .wrap-instructor-info {
+    display: flex !important;
+    flex-wrap: wrap;
+    gap: 10px;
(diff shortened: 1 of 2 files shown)


zamanafzal (approved):
LGTM 👍🏻 . I've tested the changes, and they were working as stated in the PR.

Image

PR #804, opened 2026-05-21
feat: improve Canvas grade sync and add test coverage

xitij2000:

This PR enhances the Canvas grade synchronization functionality with improved validation logic and
comprehensive test coverage:

**Key Changes:**

1. **Assignment Due Date Validation** - Added a new validation check to prevent grade
   synchronization for assignments that have already passed their due date. When attempting to sync
   grades for an overdue assignment, the system will now log a warning and skip the synchronization.

2. **Refactored Task Structure** - Split the `sync_user_grade_with_canvas` shared task into two
   components:
    - `sync_user_grade_with_canvas()` - The Celery shared task decorator
    - `_sync_user_grade_with_canvas()` - The internal implementation function

   This separation improves testability and allows the implementation logic to be tested
   independently without Celery.

3. **New Test Suite** - Added a new test module (`tests/test_tasks.py`) with 8 test cases
   covering:
    - Successful grade synchronization
    - Missing Canvas course IDs
    - Assignments not present in Canvas
    - Past due date scenarios
    - Missing Canvas student IDs
    - Grade lookup failures
    - API response error handling

4. **Test Configuration** - Added pytest configuration:
    - `conftest.py` - Test setup with Django settings initialization for bulk email configuration
    - `pyproject.toml` pytest settings - Configured with database reuse and deprecation warning
      filters

#### Unit Tests

Run the new test suite using the standard test process. Or you can mount this pacakge in the Tutor
openedx container, cd into the directory and run pytest.

#### Manual Testing

To manually test the functionality:

1. Create a grade record in Open edX for a Canvas-integrated course
2. Verify that grades for current assignments sync successfully with Canvas. You can do this through the shell:
   ```python
      from ol_openedx_canvas_integration.tasks import _sync_user_grade_with_canvas
      from lms.djangoapps.grades.models import PersistentSubsectionGrade
      # Look up a grade record's id
      _sync_user_grade_with_canvas(grade_id)
        
   ```
3. Set an assignment's due date to the past in Canvas and trigger a grade sync - confirm it's skipped with a
   warning log

### DIFF
--- a/src/ol_openedx_canvas_integration/pyproject.toml
+++ b/src/ol_openedx_canvas_integration/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "ol-openedx-canvas-integration"
-version = "0.7.1"
+version = "0.7.2"
 description = "An Open edX plugin to add canvas integration support"
 authors = [
   {name = "MIT Office of Digital Learning"}
--- a/src/ol_openedx_canvas_integration/tests/test_tasks.py
+++ b/src/ol_openedx_canvas_integration/tests/test_tasks.py
@@ -1,6 +1,6 @@
 """Tests for Canvas integration tasks"""
 
-from datetime import datetime, timedelta
+from datetime import UTC, datetime, timedelta
 from unittest.mock import MagicMock, patch
 
 import requests
@@ -39,7 +39,6 @@ class TestSyncUserGradeWithCanvas(TestCase):
         self.user = USER_MODEL.objects.create_user(
             username="student",
             email="student@example.com",
-            password="password",  # noqa: S106 # pragma: allowlist secret
         )
         self.grade_instance = PersistentSubsectionGrade.update_or_create_grade(
             user_id=self.user.id,
@@ -71,12 +70,14 @@ class TestSyncUserGradeWithCanvas(TestCase):
     ):
         """Test successful grade sync to Canvas"""
         mock_client = mock_client_class.return_value
+        # The graded assignment is linked to canvas
         mock_client.get_canvas_assignments.return_value = {
             str(self.usage_key): {
                 "id": self.canvas_assignment_id,
                 "due_at": "",
             }
         }
+        # The user is linked to a Canvas user
         mock_client.get_student_id_by_email.return_value = self.canvas_user_id
         mock_client.update_assignment_grades.return_value = MagicMock(
             status_code=requests.codes.ok
@@ -101,7 +102,7 @@ class TestSyncUserGradeWithCanvas(TestCase):
         self,
         mock_client_class,
     ):
-
+        """Test that grades are not synced if course has no linked canvas id"""
         _sync_user_grade_with_canvas(self.grade_id)
 
         mock_client_class.assert_not_called()
@@ -115,7 +116,9 @@ class TestSyncUserGradeWithCanvas(TestCase):
         self,
         mock_client_class,
     ):
+        """Test that assignments that are not synced to Canvas are not updated."""
         mock_client = mock_client_class.return_value
+        # There are assignment synced to Canvas, but not the one being graded
         mock_client.get_canvas_assignments.return_value = {
             "dummy-key": {
                 "id": self.canvas_assignment_id,
@@ -136,9 +139,11 @@ class TestSyncUserGradeWithCanvas(TestCase):
         self,
         mock_client_class,
     ):
+        """Test that grades are not synced for assignments past due date"""
         mock_client = mock_client_class.return_value
+        # Assignment is synced to Canvas, but past due date
         mock_client.get_canvas_assignments.return_value = {
-            "dummy-key": {
+            str(self.usage_key): {
                 "id": self.canvas_assignment_id,
                 "due_at": str(datetime.now(tz=UTC) - timedelta(days=1)),
             }
@@ -157,13 +162,16 @@ class TestSyncUserGradeWithCanvas(TestCase):
         self,
         mock_client_class,
     ):
+        """Test that grades are not synced if user is not linked to a Canvas user"""
         mock_client = mock_client_class.return_value
+        # The assignment is linked to Canvas ...
         mock_client.get_canvas_assignments.return_value = {
             str(self.usage_key): {
                 "id": self.canvas_assignment_id,
                 "due_at": "",
             }
         }
+        # ... but the user is not linked to a Canvas user
         mock_client.get_student_id_by_email.return_value = None
 
         _sync_user_grade_with_canvas(self.grade_id)
@@ -215,6 +223,7 @@ class TestSyncUserGradeWithCanvas(TestCase):
         mock_get_grades,
         mock_client_class,
     ):
+        """Test that grades are not synced if the Canvas API returns an error"""
         mock_client = mock_client_class.return_value
         mock_client.get_canvas_assignments.return_value = {
             str(self.usage_key): {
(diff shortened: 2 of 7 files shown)


src/ol_openedx_canvas_integration/tests/test_tasks.py
```diff
@@ -0,0 +1,235 @@
...
+from django.test import TestCase, override_settings
+from lms.djangoapps.grades.models import PersistentSubsectionGrade
+from ol_openedx_canvas_integration.tasks import _sync_user_grade_with_canvas
+from opaque_keys.edx.keys import CourseKey, UsageKey
+from openedx.core.djangolib.testing.utils import skip_unless_lms
+
+USER_MODEL = get_user_model()
+
+
+@override_settings(BULK_EMAIL_DEFAULT_RETRY_DELAY=10, BULK_EMAIL_MAX_RETRIES=5)
```
tecoholic, line 15: These settings are added in conftest.py for compatibility and are not actually used in any of the tested functionality, right? Do we need to override them here?

src/ol_openedx_canvas_integration/tests/test_tasks.py
```diff
@@ -0,0 +1,235 @@
...
+        mock_client.get_canvas_assignments.return_value = {
+            "dummy-key": {
+                "id": self.canvas_assignment_id,
+                "due_at": "",
+            }
+        }
+
+        _sync_user_grade_with_canvas(self.grade_id)
+
+        mock_client.update_assignment_grades.assert_not_called()
```
tecoholic, line 128: Why is this not synced? It would be clear if the reason is in the test case name itself, or at least a docstring explaining it. From the test, it looks like canvas returns an assignment with no due date, which should ideally be syncing the grades. Why is it not?

src/ol_openedx_canvas_integration/tests/test_tasks.py
```diff
@@ -0,0 +1,235 @@
...
+            status_code=requests.codes.ok
+        )
+
+        grade_obj = MagicMock()
+        grade_obj.percent_graded = 0.85
+        mock_get_grades.return_value = {"dummy-key": {self.user: grade_obj}}
+
+        _sync_user_grade_with_canvas(self.grade_id)
+
+        mock_client.update_assignment_grades.assert_not_called()
```
tecoholic, line 208: Not really sure what's going on in this test. Test function name says, it's "sync error", the docstring says "test successful grade sync", finally it's asserted as "not called". What's this test actually checking?

I am guessing it's this block?

```
   try:
        grade = grade_dict[grade_instance.full_usage_key][openedx_user]
        payload = dict([update_grade_payload_kv(canvas_user_id, grade.percent_graded)])
    except (KeyError, AttributeError):
```

Can we make the docstring a bit more informative about why this condition occurs?

tecoholic (changes requested):
@xitij2000 I tested the changes with an actualy Canvas instance and Open edX instance. They work as expected and I can verify it from the logs as well.

I think we can improve the tests a bit and then it should be ready.

src/ol_openedx_canvas_integration/tests/test_tasks.py
xitij2000: Nope, this is from before i added them to the test config.

src/ol_openedx_canvas_integration/tests/test_tasks.py
xitij2000: I've updated the test. I too forgot why this shouldn't sync, so it definitely needs clarification!

src/ol_openedx_canvas_integration/tests/test_tasks.py
xitij2000: Sorry about that! I basically copied the success case and then started introducing failures to test all possible failures which left in docstring. I've updated this. 

To be honest, I don't know why this would occur. This code is only run if the a grade is generated, so why would it not be returned from here, I'm not sure of the whole grading mechanism to know why we'd generate an object but still have the grading API not return a grade. If you know why this could happen, do tell and I'll update the docstring.

tecoholic: I am 100% on this. I will try with what I know.

The grade calculation is bit complex as the "grade_percent" is calculated during runtime. So even though we get a `PersistantGrade` ID, I remember choosing to just call the LMS's Python API and get it to return the grades.

I think this can sometime return no results. I am not sure why exactly - maybe there are optional problems or something.

And secondly, we just reused the [get_subsection_user_grades](https://github.com/open-craft/open-edx-plugins/blob/kshitij/grade-sync-fix/src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/api.py#L79-L79) function by introducing optional parameters to filtering and reused the existing logic. Since it applied additional filtering, we could potentially have empty dicts.

tecoholic (approved):
@xitij2000 :+1:

- [x] I tested this: Changed the due data on Canvas and verified that the grades aren't synced past the due date.
- [x] I read through the code
- [ ] I checked for accessibility issues - *NA*
- [x] Includes documentation

xitij2000:
The tests are failing here because I added support for testing this codebase but the fixes for the old tests is in the other pr. I'll add them here.

src/ol_openedx_canvas_integration/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "ol-openedx-canvas-integration"
-version = "0.6.0"
+version = "0.7.1"
```
arslanashraf7: Let's call it `0.6.1`
```suggestion
version = "0.6.1"
```

src/ol_openedx_canvas_integration/tests/test_tasks.py
```diff
@@ -0,0 +1,247 @@
...
+        )
+        self.canvas_course_id = "canvas-123"
+        self.email = "student@example.com"
+        self.canvas_user_id = 456
+        self.canvas_assignment_id = 789
+        self.user = USER_MODEL.objects.create_user(
+            username="student",
+            email="student@example.com",
+            password="password",  # pragma: allowlist secret # noqa: S106
+        )
```
arslanashraf7: Could we use `from common.djangoapps.student.tests.factories import UserFactory` instead?

arslanashraf7 (changes requested):
Along with some inline comments, this is the behavior I noticed:

1. The code works fine in the scenario where a user submits a response to a problem on an edX instance; i.e., the log prints that the grade sync was skipped because the due date was in the past.
2. _However,_ there is a second case where the instructor has the option to push the grades from the `Instructor Dashboard->Canvas->Push all MITx grades to Canvas` (Screenshot below), and in this case the grades still get synced even if the due date was in the past. I think we should handle this case as well.

**Screenshot:**

Image

arslanashraf7 (commented):
Along with the above feedback, It would be nice to add something in the Readme section about this new conditional in the grades syncing.

src/ol_openedx_canvas_integration/pyproject.toml
xitij2000: @arslanashraf7 I put it as 7.1 since there is another PR I made the bumps the version to 0.7. I'll update this based on whichever merges first.

arslanashraf7 (approved):
This one looks good 👍, but if you want to merge it before https://github.com/mitodl/open-edx-plugins/pull/788 then the version will need to be changed in `pyproject.toml`. Also, There are some failing tests which we need to fix before this can be merged.

xitij2000:
@arslanashraf7 I've updated this PR after the recent merge.

arslanashraf7:
@xitij2000 Could you please rebase this PR again? You'll need to bump the version too. I just merged https://github.com/mitodl/open-edx-plugins/pull/816, which fixes a bug that was introduced in https://github.com/mitodl/open-edx-plugins/pull/788.

xitij2000:
@arslanashraf7 Done

arslanashraf7 (approved):
👍

PR #809, opened 2026-06-05
feat: update perms for course lang view

asadali145:

https://github.com/mitodl/hq/issues/11650

Remove auth required permission from the API for course language. Adds basic throttling for authenticated and unauthenticate users.

### How can this be tested?

- Hit http://local.openedx.io:8000/course-translations/api/course-language/<COURSE_KEY> with both authenticated and unauthenticated users. Response should be course language code with 200.
- Try to hit more than 20 requests per minute and see the rate limit error.

### DIFF
--- a/src/ol_openedx_auto_select_language/ol_openedx_auto_select_language/views.py
+++ b/src/ol_openedx_auto_select_language/ol_openedx_auto_select_language/views.py
@@ -18,6 +18,18 @@ from ol_openedx_auto_select_language.utils import LanguageCode
 log = logging.getLogger(__name__)
 
 
+class CourseLanguageAnonRateThrottle(AnonRateThrottle):
+    """Throttle anonymous requests for course language endpoint."""
+
+    rate = "20/min"
+
+
+class CourseLanguageUserRateThrottle(UserRateThrottle):
+    """Throttle authenticated requests for course language endpoint."""
+
+    rate = "20/min"
+
+
 class CourseLanguageView(APIView):
     """
     API View to retrieve the language of a specified course.
@@ -47,7 +59,7 @@ class CourseLanguageView(APIView):
     """
 
     permission_classes = [AllowAny]
-    throttle_classes = [AnonRateThrottle, UserRateThrottle]
+    throttle_classes = [CourseLanguageAnonRateThrottle, CourseLanguageUserRateThrottle]
 
     def get(self, request, course_key_string):  # noqa: ARG002
         """
--- a/src/ol_openedx_auto_select_language/ol_openedx_auto_select_language/views.py
+++ b/src/ol_openedx_auto_select_language/ol_openedx_auto_select_language/views.py
@@ -12,6 +12,7 @@ from rest_framework.permissions import AllowAny
 from rest_framework.response import Response
 from rest_framework.throttling import AnonRateThrottle, UserRateThrottle
 from rest_framework.views import APIView
+from django.conf import settings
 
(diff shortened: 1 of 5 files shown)


src/ol_openedx_auto_select_language/ol_openedx_auto_select_language/views.py
```diff
@@ -8,15 +8,30 @@
...
 from ol_openedx_auto_select_language.utils import LanguageCode
 
 log = logging.getLogger(__name__)
 
 
+class CourseLanguageAnonRateThrottle(AnonRateThrottle):
+    """Throttle anonymous requests for course language endpoint."""
+
+    scope = "course_language_anon"
+    rate = "20/min"
```
arslanashraf7: 20 might be too little, I think, considering the people accessing multiple UAI modules in different languages. The least we should do is make this configurable. Can we make it a setting so that we can change the value of this without a new plugin release and deployment?

src/ol_openedx_auto_select_language/tests/test_views.py
```diff
@@ -1,13 +1,31 @@
...
 from opaque_keys import InvalidKeyError
 from rest_framework import status
+from rest_framework.permissions import AllowAny
 
 MODULE = "ol_openedx_auto_select_language.views"
 
 
+def test_view_is_public_and_throttled():
+    """Test view uses public access with basic DRF throttling."""
+    assert CourseLanguageView.permission_classes == [AllowAny]
```
arslanashraf7: A better or additional assertion would be to hit this API both with a user and without a user/anon and see the response coming in properly.
The throttling assertion could be a separate unit test.

src/ol_openedx_auto_select_language/ol_openedx_auto_select_language/views.py
```diff
@@ -8,15 +8,30 @@
...
+
+
+class CourseLanguageUserRateThrottle(UserRateThrottle):
+    """Throttle authenticated requests for course language endpoint."""
+
+    scope = "course_language_user"
+    rate = "20/min"
```
arslanashraf7: same as above comment

arslanashraf7 (changes requested):
This is just the code review; I'll test it on the next pass. A couple of small changes.

src/ol_openedx_auto_select_language/ol_openedx_auto_select_language/views.py
asadali145: 20/min was quite generous as it meant allowing to switch a course every 3 seconds. I can increase it to 30/min.

arslanashraf7 (commented):
This looks good. Just one more thing I think we should be able to do. There is a [TODO](https://github.com/mitodl/open-edx-plugins/blob/main/src/ol_openedx_auto_select_language/ol_openedx_auto_select_language/urls.py#L16-L22) where we retained the older API endpoint for legacy support. I think we should remove that as well.

asadali145:
> This looks good. Just one more thing I think we should be able to do. There is a [TODO](https://github.com/mitodl/open-edx-plugins/blob/main/src/ol_openedx_auto_select_language/ol_openedx_auto_select_language/urls.py#L16-L22) where we retained the older API endpoint for legacy support. I think we should remove that as well.

done in 38dbb7a302483e9ce3e450ebeb9096905f63c3e2

arslanashraf7 (approved):
LGTM 👍

PR #810, opened 2026-06-05
feat: remove deepl and translate updates

asadali145:

https://github.com/mitodl/hq/issues/11651 and https://github.com/mitodl/hq/issues/10805#issuecomment-4430005661

Removes support for DeepL as translation provider and adds changes to translate course updates.

### How can this be tested?

- Make sure that DeepL support is remove and no code mentions DeepL. All relevant code blocks for DeepL should be removed.
- Translate a course with updates using any of the LLMs like OpenAI and confirm that the course updates are translated.


arslanashraf7 (approved):
LGTM 👍 Should we remove DeepL from `ol_openedx_ai_static_translations` as well?

asadali145:
> LGTM 👍 Should we remove DeepL from `ol_openedx_ai_static_translations` as well?

We don't support DeepL translations for static content, I think. I think there are a couple of constants related to DeepL that I'll remove in a small follow-up PR.

PR #812, opened 2026-06-10
fix(ol-openedx-git-auto-export): correct migrate_giturl task kwarg and add tests

blarghmatey:
N/A

The `migrate_giturl` management command dispatched the `async_create_github_repo` Celery task with a stale `export_course=True` keyword argument. That parameter was renamed to `export_content` when the plugin was generalized from courses to arbitrary learning contexts (libraries included), but the command was never updated to match.

Because `.s(...)` builds a Celery signature lazily without validating kwargs, the mismatch only surfaced at `job.apply_async()` dispatch time, deep in Celery's `check_arguments`, raising:

```
TypeError: async_create_github_repo() got an unexpected keyword argument 'export_course'
```

This only triggers when the command has repositories to create (courses with duplicate or missing git URLs), which is why ordinary runs didn't hit it.

This PR:

- Fixes the kwarg in `migrate_giturl.py` (`export_course` → `export_content`).
- Fixes a possibly-unbound loop variable (`course.id`) in the repository-creation progress message; it now reports the count of queued courses instead of leaking the last loop iteration's value (which would also `NameError` on an empty course list).
- Adds the plugin's first test suite (`tests/test_migrate_giturl.py`), including a regression test that builds the real Celery signatures the command dispatches and binds them against the actual task signature — so reintroducing a bad kwarg fails the test without needing a broker.
- Bumps the package version `0.8.1` → `0.8.2` and adds a `CHANGELOG.rst` (also included in the sdist packaging).

### How can this be tested?
The plugin's tests require the Open edX environment (they import `CourseOverview`, `modulestore`, and `cms.djangoapps...`), so run them inside a Tutor/edx-platform container:

```bash
./run_edx_integration_tests.sh --plugin ol_openedx_git_auto_export --skip-build
```

To validate the original bug manually: run `python manage.py migrate_giturl --all` against a deployment that has at least one course needing a new repository (duplicate or missing git URL). On `main` this raises the `TypeError` above at `apply_async`; with this change the task dispatches normally.

### DIFF
--- a/src/ol_openedx_git_auto_export/CHANGELOG.rst
+++ b/src/ol_openedx_git_auto_export/CHANGELOG.rst
@@ -1,0 +1,12 @@
+Change Log
+==========
+
+Version 0.8.2 (2026-06-10)
+---------------------------
+
+* Fixed ``migrate_giturl`` management command passing a stale ``export_course``
+  keyword to ``async_create_github_repo``, which raised ``TypeError`` during
+  parallel repository creation.
+* Fixed ``migrate_giturl`` reporting a possibly-unbound loop variable in its
+  repository-creation progress message; it now reports the count of courses.
+* Added tests for the ``migrate_giturl`` command.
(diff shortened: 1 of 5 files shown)


asadali145:
@blarghmatey is this ready for a review?

PR #816, opened 2026-06-17
fix: canvas integration template path

arslanashraf7:
None

In https://github.com/mitodl/open-edx-plugins/pull/788/ we moved the settings files inside the `lms/cms` directories, respectively, which broke the template path for the Canvas Integration tab in the instructor dashboard. This PR adds another level of directory in the path because of this file movement in the respective directories.

Screenshot 2026-06-17 at 4 25 14 PM

- [ ] Desktop screenshots

### How can this be tested?
Make sure the `Canvas` tab shows in the Instructor Dashboard.

### DIFF
--- a/src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/settings/lms/production.py
+++ b/src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/settings/lms/production.py
@@ -2,7 +2,7 @@
 
 from path import Path as path  # noqa: N813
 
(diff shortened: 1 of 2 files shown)


Anas12091101 (approved):
LGTM 👍 

We should bump the version as well.

arslanashraf7:
> LGTM 👍
> 
> We should bump the version as well.

I was planning on getting it merged with https://github.com/mitodl/open-edx-plugins/pull/804. But maybe I should bump the version; there is nothing specific for not doing it otherwise. I'll bump it

PR #823, opened 2026-07-03
fix: register Open edX filters in production settings so they survive the config override

Anas12091101:
## Problem

`lms/envs/production.py` rebuilds settings from the deployment YAML with:

```python
_YAML_TOKENS = yaml.safe_load(f)
vars().update({key: value for key, value in _YAML_TOKENS.items() if key not in [...]})
```

This runs **after** `from .common import *`, so it **wholesale-overwrites** `OPEN_EDX_FILTERS_CONFIG` with the deployment's value. Any filter pipeline step a plugin registers **only in its common settings** is therefore dropped in production — it works locally (Tutor's YAML doesn't set the key) but silently disappears in production unless the deployment hardcodes the step in its own `OPEN_EDX_FILTERS_CONFIG`.

This is why the instructor-dashboard **Canvas** / **Rapid Responses** tabs did not appear in production, and why `ol-openedx-chat-xblock` / `ol-openedx-auto-select-language` only work because ol-infrastructure lists their filters in the configmap.

## Fix

Register each plugin's pipeline step from **production settings** as well. Production plugin settings run last (via `add_plugins(SettingsType.PRODUCTION)`, after the YAML override), so the step is merged back in. Each plugin now shares an idempotent, merge-not-clobber helper between its common and production settings; the merge preserves any other configured filters/steps.

This makes the filters **self-contained** — deployments no longer need to wire them into `OPEN_EDX_FILTERS_CONFIG` themselves.

## Plugins (all patch bumps)

| Plugin | Version | Filter |
|---|---|---|
| `ol-openedx-canvas-integration` | 0.8.0 → **0.8.1** | Canvas instructor-dashboard tab |
| `ol-openedx-rapid-response-reports` | 0.5.0 → **0.5.1** | Rapid Responses tab |
| `ol-openedx-chat-xblock` | 0.4.5 → **0.4.6** | `DisableMathJaxForOLChatBlock` render filter (+ declare PRODUCTION settings in `apps.py`) |
| `ol-openedx-auto-select-language` | 0.1.1 → **0.1.2** | `AddDestLangForVideoBlock` render filter (+ declare PRODUCTION settings in `apps.py`) |

## Verification

- Simulated the exact production sequence (common register → YAML `vars().update()` clobber → production `add_plugins`) using the actual helper code, in two scenarios — ol-infra keeps the xblock filters in the configmap, and ol-infra drops them entirely. Both restore/preserve all four steps with **no duplicates**.
- `ruff check` + `ruff format --check` (0.15.17), `py_compile`, and `rstcheck --report-level=WARNING` all pass.

## Follow-up

Once released, ol-infrastructure can drop `ol_openedx_auto_select_language` / `ol_openedx_chat_xblock` from the LMS configmap's `OPEN_EDX_FILTERS_CONFIG` (they're now self-registered). The lehrer pip pins for canvas/rapid will be bumped to `0.8.1`/`0.5.1` in a companion PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


blarghmatey (approved):

blarghmatey (approved):

src/ol_openedx_auto_select_language/ol_openedx_auto_select_language/settings/lms.py
```diff
@@ -13,26 +16,5 @@ def plugin_settings(settings):
...
-            existing_filters[filter_name]["pipeline"].extend(config.get("pipeline", []))
-            # do not override fail_silently
-            if "fail_silently" in config:
-                existing_filters[filter_name].setdefault(
-                    "fail_silently", config["fail_silently"]
-                )
-
-    settings.OPEN_EDX_FILTERS_CONFIG = existing_filters
+    # Register the video-language render filter (adds to any existing pipeline).
+    register_video_language_filter(settings)
```
asadali145: Do we need to register it here if we are doing in production.py?

src/ol_openedx_canvas_integration/CHANGELOG.rst
asadali145, line 24: Does this section update after releases? Doesn't look like.

Anas12091101: The **Unreleased** section is actually empty. It's the **0.8.1** release section that you're referring to, and it looks correct to me. Doesn't need any update after the release.

src/ol_openedx_auto_select_language/ol_openedx_auto_select_language/settings/lms.py
Anas12091101: There's no need for it, but it also wouldn't cause any harm. That said, I've removed it anyway in https://github.com/mitodl/open-edx-plugins/pull/823/commits/f186ffda29d57e6f66cd60e893eb97f93afa2158.

src/ol_openedx_canvas_integration/CHANGELOG.rst
asadali145: I am talking about these points under 0.8.1:

```
* Corrected documentation for due date syncing direction (Canvas to Open edX).
* Updated management command instructions for Tutor.
* Fixed the Tutor patch example in README.rst.
```

src/ol_openedx_auto_select_language/ol_openedx_auto_select_language/settings/lms.py
asadali145, line 1: Can we rename this to common.py, the initial purpose was a bit different but I think we can name it common.py to be consistent with production.py.

asadali145 (commented):
I have couple comments, I am testing it now.

asadali145 (approved):
LGTM!

src/ol_openedx_auto_select_language/ol_openedx_auto_select_language/settings/lms.py
asadali145: Lets keep this as is for now.

PR #826, opened 2026-07-09
fix: fix race condition where library is not in the store yet

marslanabdulrauf:
https://github.com/mitodl/hq/issues/12267

This pull request improves the reliability of exporting courses and libraries to GitHub by introducing better error handling for cases where content is not yet available in the database due to transaction timing. It adds a custom exception, ensures Celery tasks are dispatched only after database commits, and updates error handling in utility and task functions.

### How can this be tested?
1. Make sure it setup properly and configurations are correct
2. Creating a new library should always create new repository
3. Library repository's readme should have correct studio url

### DIFF
--- a/src/ol_openedx_git_auto_export/ol_openedx_git_auto_export/tasks.py
+++ b/src/ol_openedx_git_auto_export/ol_openedx_git_auto_export/tasks.py
@@ -13,6 +13,7 @@ from django.conf import settings
 from opaque_keys.edx.keys import LearningContextKey
 from rest_framework import status
 
+from ol_openedx_git_auto_export.exceptions import ContentNotFoundError
 from ol_openedx_git_auto_export.models import ContentGitRepository
 from ol_openedx_git_auto_export.utils import (
     clear_stale_git_lock,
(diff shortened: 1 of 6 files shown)


arslanashraf7:
@marslanabdulrauf is this ready for review? Could you please add the `Needs review` label to the PR if so?

src/ol_openedx_git_auto_export/ol_openedx_git_auto_export/tasks.py
```diff
@@ -129,7 +142,7 @@ def async_create_github_repo(self, context_key_str, export_content=False):  # no
         return False, response_msg
 
     # Determine URL path based on content type
-    url_path = f"{content_info['content_type']}/{context_key_str}"
+    url_path = f"authoring/{content_info['content_type']}/{context_key_str}"
```
arslanashraf7: My understanding is that this is only for informational purposes. Or is there more to it? 

nit because this impacts only repo description change in my opinion: But in any case, I think this should be configurable because this value comes from infra, and it can change over time. So, making it configurable would allow us to keep up to date with the original value in infra without any code change and a new plugin deployment. This could be a new setting that defaults to `authoring`.

arslanashraf7 (approved):
LGTM with a small comment 👍

src/ol_openedx_git_auto_export/ol_openedx_git_auto_export/tasks.py
marslanabdulrauf: Yes it is only for informational purposes.

`/authoring` is usually hardcoded in `ol-infrastructure` as well, no variable there.

image

arslanashraf7: Yes, that's True. But imagine this scenario:

If they change the `/authoring` to something like `/author`. What will we need? 1) We will need to update this plugin code and change this line. 2) Deploy a new plugin on PyPI. 3) Change the plugin pin in infra. 4) Rebuild the Open edX instance deployment.

Whereas if this were a configurable setting in the plugin, we would need only: 1) update the config value to `/author` in infra; 2) rebuild the Open edX instance.

That being said, this is still a nit, so it's up to you to make this change or merge as is.

PR #833, opened 2026-07-24
feat(ol_openedx_uai_content_customization): add gsheet link support

asadali145:

None

Enhance the use of the uai content customization command by adding support for a Google Sheets URL. It still supports the CSV file.

### How can this be tested?

- Setup plugin and use a google sheet to run the command:
```
    python manage.py cms generate_uai_course_versions \\
        --processed-videos-csv "https://docs.google.com/spreadsheets/d/<ID>/edit#gid=0" \\
        [--username studio_worker] \\
        [--dry-run]

```

### DIFF
--- a/src/ol_openedx_uai_content_customization/tests/test_csv_utils.py
+++ b/src/ol_openedx_uai_content_customization/tests/test_csv_utils.py
@@ -145,28 +145,26 @@ def test_is_google_sheets_url(source, expected):
     assert is_google_sheets_url(source) is expected
 
 
-def test_fetch_csv_text_direct_url_is_not_rewritten():
+def test_fetch_csv_text_non_google_sheets_url_raises():
     """
-    A non-Google direct CSV URL is fetched as-is, without raising ValueError.
+    A non-Google URL is rejected, even one shaped like a CSV export link.
 
     This is the case even when the URL happens to contain "/export" or
     "output=csv", proving such URLs are never mistaken for Google Sheets
     export links (which would let an operator target arbitrary hosts under
-    the guise of Google Sheets support).
+    the guise of Google Sheets support). Only ``docs.google.com`` Sheets
+    links are accepted.
     """
-    mock_response = mock.Mock()
-    mock_response.text = "a,b\n1,2\n"
-    mock_response.raise_for_status = mock.Mock()
-
     direct_url = "https://example.com/export?output=csv"
-    with mock.patch(
-        "ol_openedx_uai_content_customization.csv_utils.requests.get",
-        return_value=mock_response,
-    ) as mock_get:
-        text = fetch_csv_text(direct_url)
+    with (
+        mock.patch(
+            "ol_openedx_uai_content_customization.csv_utils.requests.get"
+        ) as mock_get,
+        pytest.raises(ValueError, match=r"only docs\.google\.com Google Sheets"),
+    ):
+        fetch_csv_text(direct_url)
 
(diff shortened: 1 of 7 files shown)


src/ol_openedx_uai_content_customization/tests/test_csv_utils.py
```diff
@@ -79,6 +86,127 @@ def test_resolve_duration_code_unknown_raises():
...
+            "https://docs.google.com/spreadsheets/d/abc123/export?format=csv&gid=9",
+        ),
+    ],
+)
+def test_build_google_sheet_csv_export_url(sheet_url, expected):
+    """Share/edit links are converted to CSV export links; export links pass through."""
+    assert build_google_sheet_csv_export_url(sheet_url) == expected
+
+
+def test_build_google_sheet_csv_export_url_invalid_raises():
```
arslanashraf7, line 125: I think these can be combined.

src/ol_openedx_uai_content_customization/tests/test_csv_utils.py
```diff
@@ -79,6 +86,127 @@ def test_resolve_duration_code_unknown_raises():
...
+        ("https://evil.com/export?output=csv", False),
+        ("/path/to/file.csv", False),
+    ],
+)
+def test_is_google_sheets_url(source, expected):
+    """Only docs.google.com Sheets links are recognised — no other host."""
+    assert is_google_sheets_url(source) is expected
+
+
+def test_fetch_csv_text_direct_url_is_not_rewritten():
```
arslanashraf7: Should the non-google URLs having a similar pattern work?

src/ol_openedx_uai_content_customization/tests/test_csv_utils.py
```diff
@@ -79,6 +86,127 @@ def test_resolve_duration_code_unknown_raises():
...
+
+    mock_get.assert_called_once_with(
+        "https://docs.google.com/spreadsheets/d/abc123/export?format=csv&gid=0",
+        timeout=30,
+    )
+    assert fieldnames == ["video_file_name", "edx_video_id"]
+    assert rows == [{"video_file_name": "v004_h264.mp4", "edx_video_id": "abc-123"}]
+
+
+def test_parse_csv_from_google_sheet_url_raises_on_http_error():
```
arslanashraf7, line 199: This can be combined into a single test with params above.

arslanashraf7 (changes requested):
Works well in testing but I left some inline comments.

src/ol_openedx_uai_content_customization/tests/test_csv_utils.py
asadali145: Both these tests are for different cases, first one tests if the valid URLs are converted to CSV export URLs, second one tests for bad URLs. I think we can keep them separate.

src/ol_openedx_uai_content_customization/tests/test_csv_utils.py
asadali145: Keep these as is.

src/ol_openedx_uai_content_customization/tests/test_csv_utils.py
asadali145: I have updated the PR to disallow non-google sheets URLs.

arslanashraf7 (approved):
👍

PR #843, opened 2026-08-05
fix(canvas-integration): only queue grade sync for Canvas-linked courses

arslanashraf7:
### Related ticket
None, Related to [This Slack Thread](https://mitodl.slack.com/archives/C02649X2P1V/p1783689691874519)

## Description
- `sync_user_grade_with_canvas` was being queued as a Celery task for **every** subsection-grade save on the entire platform, even for courses with no Canvas integration — the only Canvas-linkage check happened inside the task itself, after it was already queued and picked up by a worker.
- The `post_save` receiver (`update_grade_in_canvas`) now checks Canvas linkage *before* deciding whether to queue the task, via a new cached helper `get_cached_canvas_course_id()` (60s TTL by default, configurable via `CANVAS_COURSE_ID_CACHE_TIMEOUT`) that caches both "linked" and "not linked" outcomes, so non-Canvas courses stop enqueuing the task at all instead of just short-circuiting inside it.
- The lookup is fully guarded against failures (cache backend outage, missing course, etc.) so nothing can propagate out of the signal receiver into whatever code path saved the grade.
- Documented the setting and a caveat: a grade saved in the ~60s right after linking a course to Canvas may not auto-sync (cached negative result); the plugin's existing "Push all MITx grades to Canvas" instructor action is the documented recovery path.
- Version bumped 0.8.2 → 0.8.3 with a changelog entry.

## Testing Instructions
-  Setup a canvas Course 
- Make sure that there is a `"canvas_id": <value>` set in `Other course settings` in Advanced course settings in Studio
- Now submit any grade and the task `sync_user_grade_with_canvas` should be called.
-  Now test it on a course where the `canvas_id` is not set. The task `sync_user_grade_with_canvas` should not be called. (This is the step that you should reproduce first on main branch in order to test because on main branch this task sync_user_grade_with_canvas will be called for all the courses where a grade is submitted)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/receivers.py
+++ b/src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/receivers.py
@@ -3,6 +3,7 @@
 import logging
 
 from ol_openedx_canvas_integration.tasks import sync_user_grade_with_canvas
+from ol_openedx_canvas_integration.utils import get_cached_canvas_course_id
 
 log = logging.getLogger(__name__)
 
@@ -16,5 +17,12 @@ def update_grade_in_canvas(sender, instance, created, **kwargs):  # noqa: ARG001
     updates the Canvas course if the subsection is already synced to a Canvas
     course as an assignment.
     """
+    if get_cached_canvas_course_id(instance.course_id) is None:
+        log.debug(
+            "Course %s has no linked Canvas course. Skipping grade sync.",
+            instance.course_id,
+        )
+        return
+
     log.debug("Grade updated, triggering background task")
     sync_user_grade_with_canvas.delay(instance.id)
--- a/src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/settings/lms/common.py
+++ b/src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/settings/lms/common.py
@@ -5,3 +5,4 @@ def plugin_settings(settings):
     """Settings for the canvas integration plugin."""  # noqa: D401
     settings.CANVAS_ACCESS_TOKEN = None
     settings.CANVAS_BASE_URL = None
+    settings.CANVAS_COURSE_ID_CACHE_TIMEOUT = 300
--- a/src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/settings/lms/production.py
+++ b/src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/settings/lms/production.py
@@ -17,6 +17,12 @@ def plugin_settings(settings):
     settings.CANVAS_BASE_URL = settings.ENV_TOKENS.get(
         "CANVAS_BASE_URL", settings.CANVAS_BASE_URL
     )
+    settings.CANVAS_COURSE_ID_CACHE_TIMEOUT = int(
+        settings.ENV_TOKENS.get(
+            "CANVAS_COURSE_ID_CACHE_TIMEOUT", settings.CANVAS_COURSE_ID_CACHE_TIMEOUT
+        )
+        or settings.CANVAS_COURSE_ID_CACHE_TIMEOUT
+    )
 
     # Re-register the instructor-dashboard tab filter. Production overwrites
     # OPEN_EDX_FILTERS_CONFIG wholesale from the deployment YAML, dropping the
--- a/src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/utils.py
+++ b/src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/utils.py
@@ -1,5 +1,17 @@
 """Utilities for Canvas plugin"""
 
+import logging
+
+from django.conf import settings
+from django.core.cache import cache
+from django.http import Http404
+from lms.djangoapps.courseware.courses import get_course_by_id
+
+log = logging.getLogger(__name__)
+
+CANVAS_COURSE_ID_CACHE_KEY_TEMPLATE = "canvas_course_id:{course_id}"
+_MISSING = object()
+
 
 def get_canvas_course_id(course=None):
     """Get the course Id from the course settings"""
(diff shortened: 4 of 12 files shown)


src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/utils.py
```diff
@@ -11,6 +22,52 @@ def is_canvas_dates_sync_enabled(course=None):
...
+        )
+        cached_value = None
+
+    if cached_value is not None:
+        return None if cached_value == NO_CANVAS_ID_CACHE_VALUE else cached_value
+
+    try:
+        course = get_course_by_id(course_id)
+        canvas_course_id = get_canvas_course_id(course)
+    except Exception:
```
asadali145: Catching every `Exception` here (not just the "course not found" case) means a transient failure — a modulestore timeout, a DB blip — for a genuinely Canvas-linked course gets treated identically to "not linked": `get_cached_canvas_course_id` returns `None`, and `update_grade_in_canvas` silently skips the sync. Nothing is cached on this path (good, so it retries next save), but until the transient issue clears, grade sync is silently dropped every time with only a `log.warning` breadcrumb.

Consider narrowing this to the exceptions `get_course_by_id` actually raises for "not found" (e.g. `Http404`), and letting anything else either propagate with a distinct log level (`log.exception`) or at least not get conflated with the "not linked" result.

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/receivers.py
```diff
@@ -16,5 +17,12 @@ def update_grade_in_canvas(sender, instance, created, **kwargs):  # noqa: ARG001
     updates the Canvas course if the subsection is already synced to a Canvas
     course as an assignment.
     """
+    if get_cached_canvas_course_id(instance.course_id) is None:
```
asadali145, line 20: This uses an explicit `is None` check (good — avoids treating a falsy-but-valid Canvas course id as unlinked). But `tasks.py`'s own internal check (`_sync_user_grade_with_canvas`) still uses `if not canvas_course_id:`. If a Canvas course id could ever be `0` or `""`, the receiver would queue the task here but the task would then skip it anyway — so the `is None` fix doesn't actually change end-to-end behavior. Worth aligning both checks on the same semantics (probably `is None` in both places) so the intent stated in this one doesn't get silently undone downstream.

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/settings/lms/production.py
```diff
@@ -17,6 +17,9 @@ def plugin_settings(settings):
     settings.CANVAS_BASE_URL = settings.ENV_TOKENS.get(
         "CANVAS_BASE_URL", settings.CANVAS_BASE_URL
     )
+    settings.CANVAS_COURSE_ID_CACHE_TIMEOUT = settings.ENV_TOKENS.get(
+        "CANVAS_COURSE_ID_CACHE_TIMEOUT", settings.CANVAS_COURSE_ID_CACHE_TIMEOUT
```
asadali145: If an ops YAML override ever sets `CANVAS_COURSE_ID_CACHE_TIMEOUT` explicitly to `null`, `ENV_TOKENS.get(...)` returns that literal `None` (the default only kicks in when the key is absent), and `cache.set(key, value, None)` in Django means "cache forever." Combined with the "not linked" sentinel in `get_cached_canvas_course_id`, a bad/negative result computed during an outage could get permanently stuck instead of expiring in 60s. Similarly, a string value from YAML (`"60"` instead of `60`) could raise inside `cache.set`, which the blind except then silently swallows — so the cache never populates and every grade save falls back to a full course lookup, which is strictly worse than pre-PR behavior. Consider validating/coercing this value, e.g. `int(settings.ENV_TOKENS.get(...) or 60)`.

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/utils.py
```diff
@@ -11,6 +22,52 @@ def is_canvas_dates_sync_enabled(course=None):
...
+    try:
+        cached_value = cache.get(cache_key)
+    except Exception:
+        log.warning(
+            "Could not read cached Canvas course id for %s", course_id, exc_info=True
+        )
+        cached_value = None
+
+    if cached_value is not None:
+        return None if cached_value == NO_CANVAS_ID_CACHE_VALUE else cached_value
```
asadali145: Nit: the `NO_CANVAS_ID_CACHE_VALUE = "__no_canvas_id__"` sentinel works, but a miss-marker object is slightly more idiomatic and sidesteps even the theoretical case where real Canvas data collides with the sentinel string: `_MISS = object()`; `cached = cache.get(cache_key, _MISS)`; `if cached is not _MISS: return cached`, caching `None` directly for "not linked". Not a blocker.

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/settings/lms/common.py
```diff
@@ -5,3 +5,4 @@ def plugin_settings(settings):
     """Settings for the canvas integration plugin."""  # noqa: D401
     settings.CANVAS_ACCESS_TOKEN = None
     settings.CANVAS_BASE_URL = None
+    settings.CANVAS_COURSE_ID_CACHE_TIMEOUT = 60
```
asadali145: I think this is too less, canvas course ID won't change frequently. Can we bump it a higher value? Maybe atleast 5m or maybe an hour?

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/receivers.py
```diff
@@ -16,5 +17,12 @@ def update_grade_in_canvas(sender, instance, created, **kwargs):  # noqa: ARG001
...
+        log.debug(
```
asadali145, line 21: Why log.debug?

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/utils.py
```diff
@@ -11,6 +22,52 @@ def is_canvas_dates_sync_enabled(course=None):
...
+    """
+    cache_key = CANVAS_COURSE_ID_CACHE_KEY_TEMPLATE.format(course_id=course_id)
+
+    try:
+        cached_value = cache.get(cache_key)
+    except Exception:
+        log.warning(
+            "Could not read cached Canvas course id for %s", course_id, exc_info=True
+        )
+        cached_value = None
```
asadali145: Do we really need this try-except block?

src/ol_openedx_canvas_integration/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "ol-openedx-canvas-integration"
-version = "0.8.2"
+version = "0.8.3"
```
asadali145, line 3: Can you please rebase your PR? This change should trigger uv.lock update that we added in pre-commit.

asadali145 (changes requested):

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/receivers.py
arslanashraf7: `debug` is intentional here: this fires on every graded-interaction save for every course platform-wide that isn't Canvas-linked — the overwhelming majority of saves on the platform. `info`/`warning` would flood LMS logs with a steady-state, expected outcome. `debug` keeps it available for local troubleshooting without adding noise in production.

src/ol_openedx_canvas_integration/ol_openedx_canvas_integration/utils.py
arslanashraf7: Yes — this was added specifically because an earlier round of review feedback found that letting a `cache.get()` failure propagate here meant the outage either crashed the `post_save` signal receiver (breaking grade save) or, if caught too broadly upstream, got conflated with a real "not linked" result. Isolating it means a cache-read outage is treated as a plain miss (falls through to the course lookup) instead of either failure mode.

asadali145 (approved):
LGTM!

PR #848, opened 2026-08-11
ci: cache the openedx-dev image to speed up Tutor integration tests

asadali145:

None

Every `integration-tests` matrix leg rebuilt the `openedx-dev` image from scratch, spending ~13 minutes of a 25–40 minute job on work that is identical across runs. This builds the image nightly and pulls it from GHCR instead.

- **`ci-image-cache.yml` (new)** builds the image per Open edX branch and pushes it to GHCR. It sits off the critical path, so `packages:write` is never held by a job that runs pull request code, and the nightly cadence bounds how far the image's pinned dependencies can drift.
- **`setup-tutor` composite action (new)** is shared by the workflow that pushes the cache and the one that pulls it, so the two cannot disagree about the tag.
- **Pins the image build to the edx-platform branch under test** via `OPENEDX_COMMON_VERSION`. Tutor otherwise defaults to its own target branch, which baked in dependencies that did not match the source mounted at test time.
- **Resolves the Python version per branch** — `master` now requires >= 3.12 while the release branches are still on 3.11.
- **Builds the plugin packages once** in their own job and shares them by artifact, rather than three legs each producing the same `dist/`.
- **Reads the branch matrix from `.github/edx-branches.json`** so both workflows stay in step.
- **`ghcr-cleanup.yml` (new)** prunes superseded cache images.
- **`run_edx_integration_tests.sh`** installs uv with `pip` rather than the network install script, and uses `uv pip install` throughout.

If the pull misses, the job falls back to building locally, so tests stay runnable against an empty or stale cache.

### How can this be tested?

CI on this PR is the primary test — all three legs (`master`, `release/teak`, `release/ulmo`) should be green.

To confirm the caching path specifically, check a leg's logs for:

- `Pull cached openedx-dev image` reporting a hit, with `Build Tutor images` then skipped. Only `release/teak` has a cache image today, so it is the only leg expected to hit before merge — see Additional Context.
- The image's dependencies matching the mounted source. On `master` this is visible as `openedx-filters==3.8.0`; the mismatch this PR fixes showed up as `3.4.1`.

After merge, `ci-image-cache.yml` runs for the first time (nightly at 04:00 UTC, or trigger it directly via `workflow_dispatch`). Confirm it pushes an image for each branch, then confirm a later PR gets cache hits on all three legs.

### Additional Context

Replaces #802, which accumulated a long exploratory history while the caching approach was worked out. The code review thread and CI iterations there are useful background; this PR is the same net change squashed into one commit.

Deliberate tradeoffs and things worth a reviewer's attention:

- **The cache is empty for two of three branches until this merges.** `ci-image-cache.yml` cannot run until it exists on the default branch, so `master` and `release/ulmo` still pay the full local build on this PR. `release/teak` hits an image that predates this work, which also means its leg is not yet exercising the new `OPENEDX_COMMON_VERSION` pin — the first nightly after merge is what to watch. `master` and `release/ulmo` did build fresh under the pin and passed.
- **Up to a day of dependency drift remains possible.** The image bakes dependencies when it is built; tests mount the branch tip at PR time. A dependency bump landing between the two reintroduces the class of `ImportError` this PR fixes, and on a cache hit there is no fallback build to absorb it. Closing that gap fully means either putting the edx-platform SHA in the cache tag (which would eliminate `master`'s hit rate) or reinstalling requirements in-container. A re-run is the intended mitigation for now.
- **`master` structurally misses the cache between 00:00 and 04:00 UTC**, since its tag names the current date but the nightly that creates it has not run yet.
- **Two cache steps were removed rather than repaired.** The `pip` cache had no content component, so `actions/cache` skipped the save on every hit and froze the entry permanently while holding ~30 MB per leg. The `edx-platform` cache logged `Invalid pattern ... '..' is not allowed` and never saved at all — the checkout has to be a sibling of the workspace, so the `..` is unavoidable. Both guarded work measured in seconds.
- **`ghcr-cleanup.yml` keeps 15 tagged versions.** `master` adds one per day while the release branches reuse stable tags, so a long-unchanged release tag could in principle age out and be trimmed while still in use.
- **The `if [ $CI ]` block in `run_edx_integration_tests.sh` is inert** under this workflow, because `docker compose run` does not forward `CI`. It is kept with a comment rather than deleted, since it predates this PR and a hand-run `CI=1` from `/openedx/edx-platform` is still coherent.

### DIFF
--- a/.github/actions/setup-tutor/action.yml
+++ b/.github/actions/setup-tutor/action.yml
@@ -1,0 +1,109 @@
+---
+name: Set up Tutor
+description: >-
+  Install the Tutor release matching an Open edX branch and derive the
+  openedx-dev image names. Shared by ci.yml (which pulls the cached image)
+  and ci-image-cache.yml (which builds and pushes it) so the two can never
+  disagree about the cache tag.
+inputs:
+  edx_branch:
+    description: Open edX branch under test, e.g. master or release/teak
+    required: true
+  ghcr_owner:
+    description: GitHub org that owns the openedx-dev-cache package
+    required: true
+outputs:
+  branch_slug:
+    description: edx_branch with slashes replaced, safe for use in an image tag
+    value: ${{ steps.names.outputs.branch_slug }}
+  version_tag:
+    description: Installed Tutor version, sanitised for use in an image tag
+    value: ${{ steps.names.outputs.version_tag }}
+  openedx_dev_image:
+    description: Local image name Tutor expects, from DOCKER_IMAGE_OPENEDX_DEV
+    value: ${{ steps.names.outputs.openedx_dev_image }}
+  cache_image_tag:
+    description: Fully qualified GHCR tag holding the cached openedx-dev image
+    value: ${{ steps.names.outputs.cache_image_tag }}
+  tutor_root:
+    description: TUTOR_ROOT as Tutor itself reports it, holding env/
+    value: ${{ steps.names.outputs.tutor_root }}
+  compose_project:
+    description: Docker Compose project name for the Tutor dev environment
+    value: ${{ steps.names.outputs.compose_project }}
+runs:
+  using: composite
+  steps:
+  - name: Install Tutor
+    shell: bash
+    env:
+      EDX_BRANCH: ${{ inputs.edx_branch }}
+    run: |
+      if [[ "$EDX_BRANCH" == "release/teak" ]]; then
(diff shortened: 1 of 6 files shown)


asadali145:
@blarghmatey could you please do a final review of the changes in this PR? I have created it as a new PR and referenced the old PR as it contained alot of extra information.

AhtishamShahid (approved):

Issue #11, opened 2021-11-15
Refactor Instructor Tasks to be used from within the plugin


arslanashraf7:
**Related to:**
https://github.com/mitodl/open-edx-plugins/issues/7#issuecomment-956173887 & https://github.com/mitodl/open-edx-plugins/issues/7#issuecomment-956292339

**Description:**
We're using the instructor tasks API to run/track tasks of canvas integration and in order to do that, we kept a couple of changes in the edx-platfrom(https://github.com/mitodl/edx-platform/pull/274/files) while creating the plugin.

At that point while creating the plugins we couldn't see a direct way of migrating all of our plugin code from edx-platform to our plugin since the instructor dashboard doesn't support whole plugin-based tabs support. So,  we decided to migrate the possible code sections into the plugin and decided to work on the other ones later.

**What needs to be done:**
As part of this ticket, we will migrate the code parts of the instructor dashboard from edx-platform(https://github.com/mitodl/edx-platform/pull/274/files) to the plugin. 

This might need us to create our own API add support for instructor tasks used in https://github.com/mitodl/edx-platform/pull/274/files.

Issue #10, opened 2021-11-11
Refactor the usage of subsection block grades


arslanashraf7:
Related to #3 

We are currently using `CourseGradeFactory` to iterate over the enrolled students and their grades for Canvas sync purposes as it can be seen in a PR https://github.com/mitodl/open-edx-plugins/pull/8#discussion_r746444443. 

There is a probability that we can refactor our process of acquiring/syncing grades from [CourseGradeFactory]https://github.com/edx/edx-platform/blob/master/lms/djangoapps/grades/course_grade_factory.py#L22 to [grades_api](https://github.com/edx/edx-platform/blob/3152c14192dcccac79b7b7a76eba91b380eba78a/lms/djangoapps/grades/models_api.py#L78). This might help us reduce a chunk of our code that we are currently using.

PR #849, opened 2026-08-11
feat(ol_openedx_events_handler): mirror Open edX enrollments in MITx Online

Anas12091101:
## What are the relevant tickets?

mitodl/hq#12748

## Description (What does it do?)

Adds an LMS receiver for `COURSE_ENROLLMENT_CREATED` so enrollments made in Open edX — notably manual enrollments from the instructor dashboard — are mirrored in MITx Online through the existing enrollment webhook.

Enrollments that MITx Online created itself via the Open edX enrollment REST API are skipped by comparing the acting user (`crum`) against the new `ENROLLMENT_WEBHOOK_SERVICE_WORKER_USERNAME` setting, so the common path does not round-trip. If the acting user cannot be resolved, the webhook is dispatched anyway — a redundant call is idempotent, a dropped enrollment is not. The task also logs and drops 4xx responses instead of retrying them.

Needs to ship alongside:
- mitodl/mitxonline — forwards the payload `mode` to `create_local_enrollment`
- lehrer — bump `ol-openedx-events-handler` to `0.3.0` in `deployments/mit-ol/build_manifest.yaml`
- ol-infrastructure — add `ENROLLMENT_WEBHOOK_SERVICE_WORKER_USERNAME` to the mitxonline block in `edxapp/k8s_configmaps.py`, set to the username owning the `open-edx-service-worker/api-token` secret

Out of scope: unenrollment and mode-change sync (need `COURSE_ENROLLMENT_CHANGED` / `COURSE_UNENROLLMENT_COMPLETED` plus a deactivate endpoint), and backfill of already-affected learners.

### Setup

LMS settings:

```python
ENROLLMENT_WEBHOOK_URL = "http://<mitxonline>/api/openedx_webhook/enrollment/"
ENROLLMENT_WEBHOOK_ACCESS_TOKEN = "<OAuth2 token of a MITx Online admin>"
ENROLLMENT_WEBHOOK_SERVICE_WORKER_USERNAME = "<edX username owning MITx Online's service-worker token>"
CELERY_ALWAYS_EAGER = True   # dev only, no LMS worker in tutor dev
```

### End-to-end walkthrough (instructor dashboard → MITx Online)

The quickest way to see the whole thing work. Apply the settings under **Setup**
above, have MITx Online running, then:

**1. Give both sides the same course run and learners.**

In MITx Online, create (or pick) a `CourseRun` whose `courseware_id` matches a
course in Open edX, and make sure the learners exist in **both** systems with the
**same email** — the webhook matches on email.

**2. Mint the webhook token.** In MITx Online, create an OAuth2 access token
owned by a staff/superuser account and set it as `ENROLLMENT_WEBHOOK_ACCESS_TOKEN`
in the LMS. The endpoint is `IsAdminUser`, so a non-admin token returns 403.

**3. Enroll from the instructor dashboard.** As a course staff user, go to
`/courses/<course_id>/instructor` → **Membership** → *Batch Enrollment*, paste two
learner emails and click **Enroll**.

The LMS log should show one dispatch per learner:

```
Sending enrollment webhook for user '<learner>' in course '<course_id>' (mode: audit)
Successfully sent enrollment webhook ... Response status: 201
```

**4. Confirm the rows exist in MITx Online.**

```python
CourseRunEnrollment.all_objects.filter(run__courseware_id="<course_id>").values_list(
    "user__email", "enrollment_mode", "active"
)
```

Both learners should be there, `audit`, active. **This is the bug being fixed** —
before this change the enrollment existed only in Open edX.

**5. Confirm MITx Online's own enrollments are not echoed back.** Call the Open edX
enrollment API using MITx Online's *service-worker* token — the same credential it
uses in production:

```bash
curl -X POST http://<lms>/api/enrollment/v1/enrollment \
  -H "Authorization: Bearer <MITx Online's OPENEDX_SERVICE_WORKER_API_TOKEN>" \
  -H "Content-Type: application/json" \
  -d '{"user":"<a third learner>","course_details":{"course_id":"<course_id>"},"force_enrollment":true}'
```

The learner is enrolled in Open edX, but **no webhook is dispatched and no MITx
Online row is created** — correct, because in the real flow MITx Online already
has that record. Repeating the call with a different staff user's token *does*
dispatch, which proves the filter keys on identity rather than on the code path.

End state: three enrollments in Open edX, two in MITx Online. That gap is the
feature working.

### Scenarios

| # | Scenario | How | Webhook | MITx Online row |
|---|---|---|---|---|
| 1 | Automated tests | `./run_edx_integration_tests.sh --plugin ol_openedx_events_handler --skip-build` | — | 47 pass |
| 2 | Legacy instructor dashboard | `/courses/<id>/instructor` → Membership → Batch Enrollment | ✅ `mode: audit` | ✅ `audit` |
| 3 | New instructor dashboard MFE | MFE Members page, or `POST /api/instructor/v2/courses/<id>/enrollments/modify` (session auth) | ✅ | ✅ |
| 4 | MITx Online enrolls itself | `POST /api/enrollment/v1/enrollment` with the **service-worker** token | ❌ skipped | ❌ none |
| 5 | Control for #4 | Same call, a **different staff** token | ✅ | ✅ |
| 6 | No request context | `CourseEnrollment.enroll()` from a shell / mgmt command | ✅ fail-open | ✅ |
| 7 | Verified mode | `CourseEnrollment.enroll(user, key, mode='verified')` | ✅ `mode: verified` | ✅ `verified` |
| 8 | 4xx response | Stop MITx Online, then enroll | ✅ once, `Not retrying` | ❌ none |
| 9 | 408 / 429 | Covered by unit tests | ✅ retried | — |
| 10 | Rolled-back enrollment | `enroll()` inside `atomic()` that then raises | ❌ none | ❌ none |
| 11 | Receiver wired up | shell check below | — | — |

### Commands

**#3 — MFE endpoint (session auth; v2 API rejects bearer tokens)**

```bash
J=/tmp/c.txt; rm -f $J
curl -s -c $J -o /dev/null http://local.openedx.io:8000/login
CSRF=$(grep csrftoken $J | awk '{print $7}')
curl -s -b $J -c $J -X POST http://local.openedx.io:8000/api/user/v2/account/login_session/ \
  -H "X-CSRFToken: $CSRF" -H "Referer: http://local.openedx.io:8000/login" \
  --data-urlencode "email_or_username=<staff>" --data-urlencode "password=<pw>"

CSRF=$(grep csrftoken $J | awk '{print $7}')
curl -s -b $J -X POST \
  "http://local.openedx.io:8000/api/instructor/v2/courses/<course_id>/enrollments/modify" \
  -H "X-CSRFToken: $CSRF" -H "Referer: http://local.openedx.io:8000/" \
  -H "Content-Type: application/json" \
  -d '{"identifier":["learner@example.com"],"action":"enroll"}'
```

**#4 / #5 — enrollment REST API**

```bash
curl -s -X POST http://local.openedx.io:8000/api/enrollment/v1/enrollment \
  -H "Authorization: Bearer <TOKEN>" -H "Content-Type: application/json" \
  -d '{"user":"<learner>","course_details":{"course_id":"<course_id>"},"force_enrollment":true}'
```

`force_enrollment` is required when `ENABLE_MKTG_SITE=True`; MITx Online sends it too.

**#6 / #7 — direct enroll**

```python
import crum
from common.djangoapps.student.models import CourseEnrollment
crum.set_current_user(staff)          # omit for #6 (fail-open)
CourseEnrollment.enroll(learner, course_key, mode='verified')   # #7
```

**#8 — 4xx not retried**

```bash
cd ~/Desktop/mitxonline && docker compose stop web
# enroll from the dashboard, then:
docker logs --since 120s <lms> 2>&1 | grep -E "rejected|Not retrying"   # expect exactly one
docker compose start web
```

**#10 — rollback protection**

```python
with transaction.atomic():
    CourseEnrollment.enroll(learner, course_key)   # event fires here
    raise RuntimeError("simulated audit-write failure")
```

No webhook, no MITx Online row. Without `transaction.on_commit` this creates a
**phantom row** — MITx Online believing the learner is enrolled when edX rolled back.

**#11 — receiver registration**

```python
from openedx_events.learning.signals import COURSE_ENROLLMENT_CREATED
print([r[1]().__name__ for r in COURSE_ENROLLMENT_CREATED.receivers if r[1]()])
# expect handle_course_enrollment_created in the list
```

### Verify / reset

```bash
# MITx Online rows
docker compose exec -T -e PYTHONPATH=/src web bash -c "cd /src && python -c \"
import django,os; os.environ.setdefault('DJANGO_SETTINGS_MODULE','main.settings'); django.setup()
from courses.models import CourseRunEnrollment
for e in CourseRunEnrollment.all_objects.filter(run__courseware_id='<course_id>'):
    print(e.user.email, e.enrollment_mode, e.active)\""
```

Reset between scenarios by deleting the `CourseEnrollment` rows in edX and the
`CourseRunEnrollment` rows in MITx Online for the course run.

### Out of scope

- Unenrollment and mode changes are not mirrored (needs `COURSE_ENROLLMENT_CHANGED`
  / `COURSE_UNENROLLMENT_COMPLETED` and a deactivate endpoint).
- No backfill of learners manually enrolled before this ships — MITx Online's
  `create_local_enrollments --runs=...` covers that.
- Mode is almost always `audit` in practice: the dashboard hardcodes the default,
  and the REST API applies `mode` *after* creation (firing `CHANGED`, not `CREATED`).


src/ol_openedx_events_handler/ol_openedx_events_handler/handlers/enrollment.py
```diff
@@ -0,0 +1,142 @@
...
+        log.warning(
+            "ENROLLMENT_WEBHOOK_SERVICE_WORKER_USERNAME is not configured. "
+            "Enrollment webhooks will be dispatched for every enrollment, "
+            "including the ones created by the webhook consumer itself."
+        )
+        return False
+
+    current_username = getattr(get_current_user(), "username", None)
+    if not current_username:
+        log.debug(
```
asadali145: Any specific reason to set this to debug log? Why not simple info?

src/ol_openedx_events_handler/ol_openedx_events_handler/handlers/enrollment.py
```diff
@@ -0,0 +1,142 @@
...
+                user_email=user_email,
+                course_key=course_key,
+                mode=mode,
+            )
+        except Exception:
+            log.exception(
+                "Failed to queue the enrollment webhook for user '%s' in course '%s'.",
+                user_email,
+                course_key,
+            )
```
asadali145, line 137: Why this try except?

src/ol_openedx_events_handler/ol_openedx_events_handler/handlers/enrollment.py
asadali145, line 1: I see quite a few unnecessary blank lines. Blank lines should enhance the code readability and in this case they are not doing that.

src/ol_openedx_events_handler/ol_openedx_events_handler/tasks.py
```diff
@@ -69,6 +78,82 @@ def notify_course_access_role_addition(user_email, course_key, role):
...
+        course_key,
+        mode,
+    )
+
+    response = requests.post(
+        webhook_url,
+        json=payload,
+        headers=headers,
+        timeout=REQUEST_TIMEOUT,
+    )
```
asadali145: I think some of the code can go to a common util that takes payload and sends request. Can used to reduce code duplication like setting headers and the request itself.

src/ol_openedx_events_handler/ol_openedx_events_handler/tasks.py
```diff
@@ -69,6 +78,82 @@ def notify_course_access_role_addition(user_email, course_key, role):
...
+    if is_client_error and response.status_code not in RETRYABLE_CLIENT_ERRORS:
+        log.error(
+            "Enrollment webhook rejected for user '%s' in course '%s'. "
+            "Response status: %s, body: %s. Not retrying.",
+            user_email,
+            course_key,
+            response.status_code,
+            response.text,
+        )
+        return
```
asadali145, line 151: What's the mechanism of retrying such enrollments?

src/ol_openedx_events_handler/ol_openedx_events_handler/tasks.py
```diff
@@ -69,6 +78,82 @@ def notify_course_access_role_addition(user_email, course_key, role):
...
+
+    response.raise_for_status()
+
+    log.info(
+        "Successfully sent enrollment webhook for user '%s' in course '%s'. "
+        "Response status: %s",
+        user_email,
```
asadali145, line 157: Should this say something like successfully enrolled?

asadali145 (commented):
This is just the code review, I have a few minor questions and suggestions.

asadali145 (changes requested):
Looks good in testing. @Anas12091101 can you please look the review comments i left earlier?

src/ol_openedx_events_handler/ol_openedx_events_handler/tasks.py
Anas12091101: 5xx / connection errors / timeouts raise via raise_for_status(), so they retry. 4xx hits the return, because it can't succeed unchanged — no MITx Online account for that email, or the course run isn't there. Those aren't recovered automatically;

src/ol_openedx_events_handler/ol_openedx_events_handler/handlers/enrollment.py
Anas12091101: Because this runs from on_commit, it fires after the enrollment is already committed and outside the signal dispatch (openedx-events uses send_robust, which would otherwise absorb it).

If .delay() raises here, the exception lands inside the transaction.atomic() block in lms/djangoapps/instructor/utils.py, and the dashboard reports "Something went wrong while processing this learner" for someone who is enrolled. Raising can't undo the enrollment or re-queue the task, so logging is the only useful thing left. I've expanded the docstring to explain this.

asadali145 (changes requested):
Unenrollment from the instructor dashboard does not revert/deactivate enrollment on MITx Online. Can we please fix it?

Anas12091101:
I think that's out of scope here. We also left the unenrollment for the instructor webhook

asadali145:
@pdpinch, should we sync unenrollments from instructor dashboard or we can skip it?

asadali145 (approved):
Approving this one, we'll do unenrollment as a followup.

PR #871, opened 2026-09-15
feat(ol_openedx_course_sync): instructor dashboard tab for problem actions

asadali145:

https://github.com/mitodl/hq/issues/13317

Course teams can reset attempts or rescore a problem across a source course and all of its synced reruns from the LMS instructor dashboard, instead of asking someone with shell access to run `sync_problem_actions`.

- Adds a **Course Sync Actions** tab to the instructor dashboard via the `InstructorDashboardTabsRequested` filter, shown only to platform staff on courses that are an active sync source.
- Adds `POST courses/<course_id>/course_sync/api/sync_problem_actions`, gated on `is_staff`, returning a per-course result row for the source course and each active target.
- Moves the reset/rescore fan-out out of the management command into `utils`, so the command and the new endpoint share one implementation. Command behaviour and output are unchanged.
- Bumps the plugin to `1.2.0` and documents the tab and the endpoint, including that `OL_OPENEDX_COURSE_SYNC_SERVICE_WORKER_USERNAME` must now be set on the LMS as well as the CMS — the tab is silently omitted without it.

The page that renders inside the tab is a separate change in lehrer: **https://github.com/mitodl/lehrer/pull/235**. Neither PR is much use alone.

<details>
<summary><b>Implementation notes</b></summary>
<br>

- **Permission**: the view uses the platform's `require_global_staff`; the pipeline step uses `GlobalStaff().has_user(user)` for the same check because the filter's `user` can be `None`.
- **Tab gating**: only shown when staff *and* `get_syncable_course_mappings(course_key)` is truthy. Target/rerun courses never get it — a problem is always identified in the source course.
- **Filter registration** lives in `settings/production.py` because `lms/envs/production.py` overwrites `OPEN_EDX_FILTERS_CONFIG` wholesale from the deployment YAML; a common-settings-only registration would be discarded.
- **Delayed imports in `utils`**: `lms.djangoapps.instructor_task` transitively breaks CMS import (`bulk_email` reads a setting Studio doesn't define), so the task API is imported inside the function, not at module level.

</details>

### Screenshots

https://github.com/user-attachments/assets/c4b566b3-c56c-49fd-8ae2-5aab6cee7a7e

Screenshot 2026-09-16 at 2 59 36 PM
Screenshot 2026-09-16 at 2 59 53 PM

### How can this be tested?

- Setup the plugin following the readme
- Create a source course and some target runs.
- Go to the instructor dashboard as per the instructions in https://github.com/mitodl/lehrer/pull/235 to source course
- Validate that you see the course sync actions tab and test it.

### Additional Context

- The tab is inert without https://github.com/mitodl/lehrer/pull/235 — server adds the tab entry, the MFE renders the page. `require_global_staff` on the endpoint is the real access control; the tab gate is presentation only.
- `deployments/mit-ol/build_manifest.yaml` in lehrer pins `ol-openedx-course-sync==1.0.1` and will need `1.2.0` once this is released.
- This plugin has no `CHANGELOG.rst`, unlike most siblings, so there's no changelog entry here. Happy to start one if wanted.


src/ol_openedx_course_sync/ol_openedx_course_sync/pipeline.py
```diff
@@ -0,0 +1,67 @@
...
+    a source course's problem and fanned out to its targets from there.
+
+    Hooks into the ``InstructorDashboardTabsRequested`` filter
+    (``org.openedx.learning.instructor.dashboard.tabs.requested.v1``).
+    """
+
+    def run_filter(self, tabs, user, course_key):
+        """Append the Course Sync tab to the instructor dashboard tab list."""
+        # GlobalStaff().has_user() is `user.is_staff`, but tolerates user being None.
+        if GlobalStaff().has_user(user) and get_syncable_course_mappings(course_key):
```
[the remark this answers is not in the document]
asadali145: Fixed in 247214a, though not quite for the stated reason — worth recording why, since the 500 depends on configuration.

`run_pipeline` catches generic `Exception`, logs it, and `continue`s when `fail_silently` is set (`openedx_filters/tooling.py`, the `except Exception` branch), and `settings/filters.py` registers this filter with `fail_silently: True`. So as registered, an `ImproperlyConfigured` here is swallowed and the dashboard renders without the tab rather than 500ing.

The gap is that the registration uses `filters_config.setdefault(...)`, so if a deployment YAML already defines this filter key with `fail_silently: False`, the merge keeps that value and the exception would propagate. Relying on that was wrong, so the step now catches `ImproperlyConfigured`, logs a warning and omits the tab. That also stops a traceback being logged on every dashboard render in a misconfigured environment.

I kept `get_syncable_course_mappings` rather than moving to a presentation-only mapping check: it also enforces the `CourseSyncOrganization` active check, and dropping that would show the tab on courses where the fan-out would not actually sync anywhere.

Covered by `test_unconfigured_plugin_omits_tab_instead_of_raising` in `tests/test_pipeline.py`.

src/ol_openedx_course_sync/ol_openedx_course_sync/utils.py
```diff
@@ -396,3 +405,112 @@ def get_all_source_courses():
...
+            results.append(
+                {
+                    "course_id": course_id_str,
+                    "action": action,
+                    "status": STATUS_FAILED,
+                    "error": str(exc),
+                }
+            )
+
+        except Exception as exc:
```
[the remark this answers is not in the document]
asadali145: Checked this and the lint step does not flag it, so I have left the handler as is.

The suppression was dropped deliberately: the handler now calls `log.exception(...)`, and BLE001 does not fire on a blind except that logs the exception. Ruff told me so directly — with the `# noqa: BLE001` still present it reported `RUF100 [*] Unused `noqa` directive (unused: BLE001)`, so keeping it would fail lint rather than satisfy it.

Verified against the pinned hook rather than a local ruff: `pre-commit run` over this diff passes `ruff` and `ruff format` (hook rev `v0.16.6`, args `--extend-ignore=D1 --fix`), which is the same configuration pre-commit.ci runs.

The failing `pre-commit.ci` check on this PR was unrelated — the `uv-lock` hook, because `uv.lock` still recorded version 1.1.0 after the bump to 1.2.0. Fixed in 247214a; `uv lock --check` is clean now.

asadali145:
@pdpinch I have this PR, in combination with https://github.com/mitodl/lehrer/pull/235 to introduce a tab for the source courses to be able to reset learner attempts or re-score learners in all target courses.

- The tab will only show up if there is an active CourseSyncMapping entry as a source course.
- It will only show up for staff users.

I have added screen recording and screenshots in the PR description.

Question: should we introduce a flag to be able to disable it per course? I wouldn't recommend as we only display the tab to staff users and in source courses.

Please let me know if you have more thoughts on it.

src/ol_openedx_course_sync/ol_openedx_course_sync/apps.py
```diff
@@ -16,6 +17,13 @@ class OLOpenEdxCourseSyncConfig(AppConfig):
     verbose_name = "Open edX Course Sync"
 
     plugin_app = {
+        PluginURLs.CONFIG: {
+            ProjectType.LMS: {
+                PluginURLs.NAMESPACE: "",
+                PluginURLs.REGEX: f"courses/{COURSE_ID_PATTERN}/course_sync/api/",
```
arslanashraf7, line 23:
```suggestion
                PluginURLs.REGEX: f"courses/{COURSE_ID_PATTERN}/course_sync_actions/api/",
```

src/ol_openedx_course_sync/ol_openedx_course_sync/pipeline.py
```diff
@@ -0,0 +1,88 @@
...
+    path component, the same way the LMS builds the built-in tab URLs) so our tab
+    links stay consistent with the standard instructor dashboard tabs instead of
+    hardcoding the ``/apps`` mount point.
+    """
+    base_url = getattr(settings, "INSTRUCTOR_MICROFRONTEND_URL", None) or ""
+    base_path = urlparse(base_url).path.rstrip("/")
+    return "/".join([base_path, str(course_key).strip("/"), tab_id])
+
+
+class AddCourseSyncInstructorTab(PipelineStep):
```
arslanashraf7, line 32: Similar changes might be required on other places as well.
```suggestion
class AddCourseSyncActionsInstructorTab(PipelineStep):
```

src/ol_openedx_course_sync/ol_openedx_course_sync/views.py
```diff
@@ -0,0 +1,67 @@
...
+    except InvalidKeyError:
+        return JsonResponse({"error": f"Invalid course key: {course_id}"}, status=400)
+
+    problem_id = request.POST.get("problem_id", "")
+    try:
+        problem_usage_key = UsageKey.from_string(problem_id)
+    except InvalidKeyError:
+        return JsonResponse(
+            {"error": f"Invalid problem usage key: {problem_id}"}, status=400
+        )
```
arslanashraf7, line 48: Should we also add an additional check - Making sure that the `OL_OPENEDX_COURSE_SYNC_SERVICE_WORKER_USERNAME` is configured or not?

src/ol_openedx_course_sync/README.rst
```diff
@@ -139,3 +159,37 @@ Rescore a problem for all learners across all synced courses:
         "block-v1:ORG+COURSE+RUN+type@problem+block@abc123" \
         --username courses_service_worker \
         --only-if-higher
+
+Problem Actions from the Instructor Dashboard (LMS)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The same actions are also available as a **Course Sync** tab in the LMS instructor
```
arslanashraf7, line 166: Course Sync rename also needed here.

pdpinch:
> It will only show up for staff users.

django staff, or course staff? Since it's limited to source courses, it probably doesn't make too much of a difference. Whatever is easier is probably fine. 

> should we introduce a flag to be able to disable it per course?

I agree. A flag is unnecessary complexity.

arslanashraf7 (approved):
Just a few comments. LGTM otherwise 👍

src/ol_openedx_course_sync/ol_openedx_course_sync/views.py
asadali145: User won't see the tab if OL_OPENEDX_COURSE_SYNC_SERVICE_WORKER_USERNAME is not configured so they can never reach this API.

